PR #4, opened 2023-07-11
Recommendations for collecting calibration data

sfmig:
After a bit of research, I collected a few recommendations on how to best collect calibration data for a stereo camera configuration.

### DIFF
--- a/stereo calibration/CalibrationDataCollection.md
+++ b/stereo calibration/CalibrationDataCollection.md
@@ -1,0 +1,69 @@
+# Collecting calibration data
+
+## Overview and rationale?
+...
+
+## Recommendations
+
+### 0. Arrangement of the cameras
+There are very nice recommendations on the arrangement of the cameras in the [StereoMorph docs](https://aaronolsen.github.io/tutorials/stereomorph/arranging_cameras_photography.html)
+- As much FOV overlap as possible? --- see StereoMorph
+- Remember! We are calibrating the cameras for a specific relative position of the cameras, zoom and focus levels!
+- Set the cameras to the smallest aperture
+
+### 1. Design of the checkerboard
+- The pattern should be on a flat and rigid surface 
+
+- White border around it, of roughly the same width as one of the cells (this makes the detection more robust in various environments see [here](https://docs.opencv.org/4.2.0/d9/d0c/group__calib3d.html#gad0e88e13cd3d410870a99927510d7f91))
+
+- Min number of **inner corners**
+    - *Why inner corners?*: "All chessboard related algorithm expects amount of inner corners as board width and height" [^2] - openCV parametrises the width and height based on number of inner corners
+    - 
+
+
+
+- Creating a calibration pattern
+    - [OpenCV](https://docs.opencv.org/4.x/da/d0d/tutorial_camera_calibration_pattern.html)
+    - [Calib.io](https://calib.io/pages/camera-calibration-pattern-generator)
+    - [Mark Hedley Jones blog post](https://markhedleyjones.com/projects/calibration-checkerboard-collection) --- not sure if these print out with a white border!
+
+### 2. Position and orientation of the checkerboard
+For a calibration frame to be useful, the checkerboard must fulfill the following criteria:
+
+- Keep the orientation of the checkboard within 90deg (conservatively 30 deg [^1]) of the "original"/default position
+    - *Why?* The algorithm takes the bottom (right?) corner of the board as the origin, and then traverses the rest of corners from there. If in some frames the board is rotated beyond 30 deg, the origin may be detected at a different physical corner, and all the corners may be traversd in a different order.
+
+- Cover several distances, and within each distance, cover (all parts of the image view)/(of the FOV overlap) [^1]
+    - *Why?*
+    - Note however that if at some distance the calibration pattern is too small, we may not be able to detect the corners reliably and the calibration may be poor. To fix this we may need a larger board (or move cameras closer to the area of interest?) 
+
+- Use a checkerboard as big as possible, ideally with of at least 8x6 squares and a square size of at least 100 mm [^1]
+    - *Why?*
+
+- For a calibration frame to be useful, the checkerboard must be fully visible in both camers
+    - *Why?* Since the algorithm first tries to search for the outer corners of the board, it will fail if the board is only partially visible. This may be avoidable using an alternative algorithm, but the approach [doesn't seem 100% robust](https://github.com/opencv/opencv/issues/15712#issuecomment-1493344373)
+
+- Number of frames captured
+    - From DLC, they claim 30-70 pairs of candidate calibration images should be sufficient, as after corner detection, some of the images might need to be discarded due to either incorrect corner detection or incorrect order of detected corners [^1]
+
+- Keep the pattern steady when aiming to collect a calibration frame
+
+## Timecode drift
+- Use approach from Anipose as validation?
+
+## Additional comments
+- Acinoset paper may be a good reference on people doing stereo in the field, in a large area
+- Should we try other patterns too (charuco boards, circles grid) ? Would it be more robust?
+
+## Useful references
+- [DeepLabCut 3D](https://deeplabcut.github.io/DeepLabCut/docs/Overviewof3D.html#jump-in-with-direct-deeplabcut-2-camera-support)
+- [Anipose]()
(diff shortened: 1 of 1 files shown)


stereo calibration/CalibrationDataCollection.md
```diff
@@ -0,0 +1,69 @@
...
+- Creating a calibration pattern
+    - [OpenCV](https://docs.opencv.org/4.x/da/d0d/tutorial_camera_calibration_pattern.html)
+    - [Calib.io](https://calib.io/pages/camera-calibration-pattern-generator)
+    - [Mark Hedley Jones blog post](https://markhedleyjones.com/projects/calibration-checkerboard-collection) --- not sure if these print out with a white border!
+
+### 2. Position and orientation of the checkerboard
+For a calibration frame to be useful, the checkerboard must fulfill the following criteria:
+
+- Keep the orientation of the checkboard within 90deg (conservatively 30 deg [^1]) of the "original"/default position
+    - *Why?* The algorithm takes the bottom (right?) corner of the board as the origin, and then traverses the rest of corners from there. If in some frames the board is rotated beyond 30 deg, the origin may be detected at a different physical corner, and all the corners may be traversd in a different order.
```
sannatitus, line 68: *traversed

stereo calibration/CalibrationDataCollection.md
```diff
@@ -0,0 +1,69 @@
...
+
+- Cover several distances, and within each distance, cover (all parts of the image view)/(of the FOV overlap) [^1]
+    - *Why?*
+    - Note however that if at some distance the calibration pattern is too small, we may not be able to detect the corners reliably and the calibration may be poor. To fix this we may need a larger board (or move cameras closer to the area of interest?)
```
sannatitus: I don't think we can move the cameras closer to the area of interest, because of topography limitations and then we would have to start a fresh traingulation, no? 
We can zoom in the lenses, perhaps. The plan for the moment is just to have the largest board possible

stereo calibration/CalibrationDataCollection.md
```diff
@@ -0,0 +1,69 @@
...
+For a calibration frame to be useful, the checkerboard must fulfill the following criteria:
+
+- Keep the orientation of the checkboard within 90deg (conservatively 30 deg [^1]) of the "original"/default position
+    - *Why?* The algorithm takes the bottom (right?) corner of the board as the origin, and then traverses the rest of corners from there. If in some frames the board is rotated beyond 30 deg, the origin may be detected at a different physical corner, and all the corners may be traversd in a different order.
+
+- Cover several distances, and within each distance, cover (all parts of the image view)/(of the FOV overlap) [^1]
+    - *Why?*
+    - Note however that if at some distance the calibration pattern is too small, we may not be able to detect the corners reliably and the calibration may be poor. To fix this we may need a larger board (or move cameras closer to the area of interest?) 
+
+- Use a checkerboard as big as possible, ideally with of at least 8x6 squares and a square size of at least 100 mm [^1]
```
sannatitus: Are we sure about 100mm? That's huge! 
If that's the case, to get 100mm 8x6 squares, we'd need 2x A2 checkerboards to fit together (over 2 meters in width!)

stereo calibration/CalibrationDataCollection.md
```diff
@@ -0,0 +1,69 @@
...
+    - *Why?*
+
+- For a calibration frame to be useful, the checkerboard must be fully visible in both camers
+    - *Why?* Since the algorithm first tries to search for the outer corners of the board, it will fail if the board is only partially visible. This may be avoidable using an alternative algorithm, but the approach [doesn't seem 100% robust](https://github.com/opencv/opencv/issues/15712#issuecomment-1493344373)
+
+- Number of frames captured
+    - From DLC, they claim 30-70 pairs of candidate calibration images should be sufficient, as after corner detection, some of the images might need to be discarded due to either incorrect corner detection or incorrect order of detected corners [^1]
+
+- Keep the pattern steady when aiming to collect a calibration frame
```
sannatitus, line 82: I assume this means physically holding the checkerboard... not traversing in the same pattern within the FOV to corner 1, then corner 2, than corner 3 and so on?

sannatitus (approved):
Looks great! Just some very minor follow-up notes. I'll slack you about the most important (100mm checkers)

stereo calibration/CalibrationDataCollection.md
```diff
@@ -1,65 +1,105 @@
...
+    - They recommend setting the cameras to the smallest aperture
+        > "The smaller the aperture, the greater the depth of field (i.e. the more things are in focus both close and far away from the camera). This is essential in a stereo camera setup because in order to digitize points accurately throughout the calibration volume they must be in focus."
+
+- Re cameras' positions:
+    - "Because the camera is often positioned half a meter or more away from the object, even a small shift of the camera can translate into a large shift in the image frame, causing large inaccuracies"
+    - Leg stoppers ([this type](https://www.amazon.co.uk/Universal-Stainless-Photography-Accessories-Replacement/dp/B0B9NXFTWP/ref=sr_1_11?keywords=tripod+rubber+feet+replacement&qid=1689006351&sr=8-11) or [these with spikes for uneven surfaces](https://www.amazon.co.uk/Universal-Non-Slip-Replacement-Compatible-Kingjoy/dp/B0B33RDMTC/ref=sr_1_12?keywords=tripod+rubber+feet+replacement&qid=1689006351&sr=8-12)) could be useful to keep the tripods as fixed as possible against the wind and on top of uneven terrain
```
sannatitus, line 26: Great idea! Our tripods aren't compatable with this, but we're getting wind shields (which will hopefully do the trick)

stereo calibration/CalibrationDataCollection.md
```diff
@@ -1,65 +1,105 @@
...
+    - "It's best to use a remote (wireless or cord) to release the shutter so you minimizing touching the shutter button on the cameras as much as possible."
```
sannatitus, line 27: Unfortunately we can't do this, as we have to write the videos onto a separate Atomos monitor which holds the SSDs (our videos are too high res & long to save on the SSD slots). We will have to push the button gently

sannatitus (approved):

stereo calibration/CalibrationDataCollection.md
sfmig: > we would have to start a fresh traingulation, no?

Yes!

Re zoom, I'd say that the larger the area of interest is in the (overlapping) field of view of both cameras, the better

stereo calibration/CalibrationDataCollection.md
sfmig: yes I agree - we discussed this offline so marking as resolved

stereo calibration/CalibrationDataCollection.md
sfmig: I don't understand this question 🤔

sannatitus: For "Keep the pattern steady when aiming to collect a calibration frame", I assume this means maintaining the orientation. For example, one should always hold it in the 8x6 configuration rather than flipping it 90 degrees to also show the 6x8 configuration. 

The other way it could be interpreted is to hold it at the same exact angle, relative to the camera, when walking around the FOV. For example, this would mean holding it perpendicular to the ground always and just walking around, keeping it steady. My understanding is that it's valuble to have a little variation with the degrees of the board relative to the camera, so when standing at one spot, one should 'tilt' the checkerboard like 10degrees to the left and 10degrees to the right. If we keep it perpendicular to the ground at all times, we won't capture minor variations in topography I think? 

Regardless, I am interpreting it as one should keep the checkerboard in the same orientation (like landscape 8x6) for all calibrations

sannatitus (approved):

stereo calibration/CalibrationDataCollection.md
sfmig: Oh I see! I actually just meant "steady" so that the frame is not blurry.

But you are right with the orientation too, as you say we want to keep it in the 8x6 / landscape configuration with small variations in all three angles of rotation (i.e., rotation around the short side of the chessboard, rotation around the long side and rotation around the axis perpendicular to the chessboard)

As you say we want to have these variations in the orientation of the chessboard in all three angles. The reason why we want this is explained a bit more [here](https://aaronolsen.github.io/tutorials/stereomorph/calibration_general.html) (very handwave-ily: it's as if we want to get as close as possible to doing a calibration with a 3D grid of points, but using a checkerboard instead, and so for that we place the board at different orientations)

PR #23, opened 2023-08-11
faster rcnn model for object detection

nikk-nikaznan:


crabs/detection_tracking/sort.py
sfmig, line 1: I was wondering if there is a more maintainable way of using SORT... I saw there is [a packaged version of SORT](https://github.com/kadirnar/sort-pip) (but there seem to be license issues?), but I think a nicer way would be to [git clone the SORT repo](https://github.com/abewley/sort/tree/master) right?

Alternatively we can check [deep SORT](https://github.com/nwojke/deep_sort) which seems like an improvement using appearance

crabs/detection_tracking/evaluate_model.py
```diff
@@ -0,0 +1,161 @@
...
+import torch
+from detection_utils import (
+    create_dataloader,
+    get_test_transform,
+    myFasterRCNNDataset,
+)
+from evaluate import test_detection
+
+# select device (whether GPU or CPU)
+device = (
```
sfmig, line 14: same comment about the scope as previously

crabs/detection_tracking/evaluate_model.py
```diff
@@ -0,0 +1,161 @@
...
+    torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+)
+
+
+class Detector_Test:
```
sfmig, line 19: same comment about using test in classes and functions. I'm not sure if pytest would be confused in this case but I still find it confusing to have the word test in something that is not an actual test?

crabs/detection_tracking/evaluate_model.py
```diff
@@ -0,0 +1,161 @@
...
+    """
+    A class for testing object detection models using pre-trained classification.
```
sfmig, line 21:
```suggestion
    A class for running object detection models using pre-trained classification.
```
maybe?

crabs/detection_tracking/evaluate_model.py
sfmig: maybe "model evaluation" or "detector evaluation"? I find I cannot tell very well what this class does from its name

crabs/detection_tracking/evaluate_model.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Load the pre-trained subject classification model.
+        """
+        # Load the pre-trained subject predictor
+        # TODO: deal with different model
+        self.trained_model = torch.load(
+            self.args.model_dir, map_location=torch.device("cpu")
+        )
+
+    def _load_dataset(self) -> None:
```
sfmig, line 64: can we reuse the equivalent code from the detector utils? Maybe we can have a parent class that the evaluation class and the detector class inherit from? Right now it feels a bit duplicated

crabs/detection_tracking/evaluate_model.py
```diff
@@ -0,0 +1,161 @@
...
+        """Load images and annotation file for training"""
+
+        self.annotation = (
+            f"{self.main_dir}/annotations/VIA_JSON_combined_coco_gen.json"
```
sfmig, line 68: same as before, it would be nice not to have this hardcoded

crabs/detection_tracking/evaluate_model.py
sfmig, line 1: looks good, but it seems we could combine a lot with the train_model module? just to avoid code duplication, needing to edit things in two places or potentially bits getting out of sync

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Get the number of input features
+    in_features = model.roi_heads.box_predictor.cls_score.in_features
+    # define a new head for the detector with required number of classes
+    model.roi_heads.box_predictor = FastRCNNPredictor(in_features, num_classes)
+
+    return model
+
+
+def train_faster_rcnn(
```
sfmig, line 48: I feel that this could be better placed in train_model.py, what do you think?

crabs/detection_tracking/models.py
sfmig, line 1: I like the idea of having a model factory kinda, maybe in the future we can add YOLO to the mix etc. Not sure this is the best way but I think it's a good starting point! We can ask Sam for software engineering advice

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def create_faster_rcnn(num_classes: int, coco_model: bool = True) -> nn.Module:
+    """Create faster rcnn model for the training
+
+    Parameters
+    ----------
+    num_classes : int
+        number of classes to train
+    coco_moder : bool
```
sfmig, line 24:
```suggestion
    coco_model : bool
```
I thought this meant whether the annotations will be in COCO format or not, but I think you mean whether to get the faster RCNN model with all the COCO classes as output, or with custom classes. Is that correct?

If so, maybe we could just have num_classes equal to None by default, and if it's None then we return the faster RCNN model with all the COCO classes.

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+        optimizer instance for the given model
+
+    Returns
+    -------
+    model : nn.Module
+        the trained model
+    """
+
+    # select device (whether GPU or CPU)
+    device = (
```
sfmig, line 74: nice! I like it here ✨ 😅

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+        torch.device("cuda")
+        if torch.cuda.is_available()
+        else torch.device("cpu")
+    )
+
+    # setup tensorboard stuff
+    layout = {
```
sfmig, line 81: nice, love being able to babysit the training with visualisation! 

I was wondering where is the MLFlow implementation? is it in a separate PR? (my understanding was that we were using MLflow to log so far, is that correct?)

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+    }
+    writer = SummaryWriter(f"/tmp/tensorboard/{int(time.time())}")
+    writer.add_custom_scalars(layout)
+
+    def log_scalar(name, value, epoch):
+        """Log a scalar value to both MLflow and TensorBoard"""
+        writer.add_scalar(name, value, epoch)
+        mlflow.log_metric(name, value)
+
+    EXPERIMENT_NAME = "baseline"
```
sfmig, line 101: should EXPERIMENT_NAME be an input to this function? I imagine we use it to identify different runs, so then hardcoding it here doesn't seem very useful right?

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+
+    try:
+        EXPERIMENT_ID = mlflow.get_experiment_by_name(
+            EXPERIMENT_NAME
+        ).experiment_id
+    except mlflow.exception.MlflowException:
+        EXPERIMENT_ID = mlflow.create_experiment(EXPERIMENT_NAME)
+
+    with mlflow.start_run(experiment_id=EXPERIMENT_ID, run_name=RUN_NAME):
+        mlflow.log_params(config)
```
sfmig, line 112: oh nice! are we using both tensorboard and MLFlow? should we just use one?

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+            for batch_idx, (imgs, annotations) in enumerate(train_dataloader):
+                i += 1
+                imgs = list(img.to(device) for img in imgs)
+                annotations = [
+                    {k: v.to(device) for k, v in t.items()}
+                    for t in annotations
+                ]
+
+                loss_dict = model(imgs, annotations)
+                losses = sum(loss for loss in loss_dict.values())
```
sfmig, line 130: I think it's slightly more standard to compute the mean loss of the batch right? (or why the sum?)

crabs/detection_tracking/models.py
```diff
@@ -0,0 +1,157 @@
...
+
+                optimizer.zero_grad()
+                losses.backward()
+                optimizer.step()
+
+            log_scalar("total_loss/train", losses, epoch)
+
+        # Upload the TensorBoard event logs as a run artifact
```
sfmig, line 138: aaah I see! so the tensorboard logs can be seen in MLFlow dashboard? nice ✨

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+from pathlib import Path
+
+import cv2
+import numpy as np
+import torch
+import torchvision.transforms as transforms
+from sort import Sort
+
+# select device (whether GPU or CPU)
+device = (
```
sfmig, line 12: same bit about the scope as before 😬

crabs/detection_tracking/inference_model.py
sfmig, line 1: I think here there is also a bit of code duplication that we could avoid... maybe this is good stuff to do pair prog with Sam and get his wisdom!

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+    torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+)
+
+
+class Detector_Inference:
+    """
+    A class for performing object detection or tracking inference on a video
+    using a pre-trained model.
```
sfmig, line 20:
```suggestion
    using a trained model.
```
why pre-trained? 🤔

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+        _load_pretrain_model(self) -> None:
+            Load the pre-trained subject classification model.
+
+        __inference(self, frame, video_file, frame_id) -> None:
+            Perform inference on a single frame of the video.
+
+        _load_video(self) -> None:
+            Load the input video and perform inference on its frames.
+
+        inference_model(self) -> None:
```
sfmig, line 43:
```suggestion
        run_inference(self) -> None:
```
maybe more clear?

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+            Perform object detection or tracking inference on the input video.
+
+    """
+
+    def __init__(self, args: argparse.Namespace) -> None:
+        self.args = args
+        self.vid_dir = args.vid_dir
+        self.score_threshold = args.score_threshold
+        self.sort_crab = Sort()
```
sfmig, line 52: what is `sort_crab`? (can it be rename to something more self-explanatory?)

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+                prediction,
+                pred_score,
+                self.score_threshold,
+                self.sort_crab,
+            )
+        return frame_out
+
+    def _load_video(self) -> None:
+        """
+        Load the input video and perform inference on its frames.
```
sfmig, line 222: maybe also include in the docstring that the multi channels are extracted if required.

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+    """
+    A class for performing object detection or tracking inference on a video
+    using a pre-trained model.
+
+    Args:
+        args (argparse.Namespace): Command-line arguments containing
+        configuration settings.
+
+    Attributes:
+        args (argparse.Namespace): The command-line arguments provided.
```
sfmig, line 27: I think some of the args are missing right?

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+                max_abs_blurred_frame,
+            )
+        )
+
+        # compute the motion channel for frame_idx
+        return np.abs(
+            background_subtracted_frame_delta - background_subtracted_frame,
+        )
+
+    def __inference(
```
sfmig, line 176: maybe rename to
```suggestion
    def run_inference_one_frame(
```
?

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+            from _inference import inference_detection
+
+            frame_out = inference_detection(
+                frame, prediction, pred_score, self.score_threshold
+            )
+
+        else:
+            from _inference import inference_tracking
+
+            frame_out = inference_tracking(
```
sfmig, line 211: I don't understand very well how we can do tracking on one frame only 🤔

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+                frame,
+                prediction,
+                pred_score,
+                self.score_threshold,
+                self.sort_crab,
+            )
+        return frame_out
+
+    def _load_video(self) -> None:
```
sfmig, line 220:
```suggestion
    def run_inference_on_video(self) -> None:
```
what do you think?

crabs/detection_tracking/inference_model.py
```diff
@@ -0,0 +1,394 @@
...
+                    mean_blurred_frame += blurred_frame
+
+                    # accumulate max absolute values
+                    max_abs_blurred_frame = np.maximum(
+                        max_abs_blurred_frame, abs(blurred_frame)
+                    )
+
+                    frame_count += 1
+            # compute the mean
+            mean_blurred_frame = mean_blurred_frame / frame_count
```
sfmig, line 278: can we reuse the utils for computing the additional channels here? To avoid code duplication and making sure we are using the same approach in training and inference.

If it's a lot for this PR feel free to open an issue

sfmig (approved):
Nice work! I did have quite a few comments (but also the PR is big 😅) 

I think going forward we should try to make smaller PRs (maybe only for the training work, then evaluation etc.) - I know it's hard tho!

I set it to "approve merging" but if you can make sure that any bits not addressed are made into issues I would really appreciate it 🙌 

Happy to chat about any of it offline if you need it!

crabs/detection_tracking/evaluate.py
```diff
@@ -0,0 +1,273 @@
+from collections import defaultdict
```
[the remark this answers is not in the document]
nikk-nikaznan: We can, but some of the utils are shared between training and testing. Will add an issue for future me to think about

crabs/detection_tracking/evaluate_model.py
nikk-nikaznan: agree - especially on the data loader section - will add to the issue

crabs/detection_tracking/evaluate_model.py
nikk-nikaznan: Yes, Sorry about that. the code used to be train and test. I changed some that failed the pytest. Will change them all to be consistent.

PR #24, opened 2023-08-15
Tests for frame extraction

### LINKED ISSUE #15: Add frame extraction tests
- [ ] test combining multiple frame extraction runs into one output subdirectory

sfmig:
Integration tests for frame extraction.

Closes #15


tests/test_integration/test_frame_extraction.py
```diff
@@ -1,51 +1,490 @@
...
+    """
+
+    return [
+        f
+        for f in Path(input_dir).glob("*")
+        if f.is_file() and not f.name.startswith(".")
+    ]
+
+
+def check_output_files(list_input_videos: list, cli_dict: dict):
```
samcunliffe, line 226:
```suggestion
def check_output_files(list_input_videos: list, cli_dict: dict) -> None:
```

tests/test_integration/test_frame_extraction.py
```diff
@@ -1,51 +1,490 @@
...
     get_list_of_sleap_videos,
 )
 
 
-@pytest.fixture(autouse=True, scope="class")
-def input_video_dir():
-    return Path(__file__).parents[1] / "data" / "clips"
+@pytest.fixture()
+def input_data_dir() -> str:
+    return str(Path(__file__).parents[1] / "data" / "clips")
```
samcunliffe, line 19: What's the benefit of a fixture (which takes three lines to define and one extra line in the argument to pytest) over a simple file-scope global?
```python
INPUT_DATA_DIR = str(Path(__file__).parents[1] / "data" / "clips")
```

samcunliffe: Hmm [the internet says this](https://stackoverflow.com/questions/75863152/pytest-fixture-vs-global-variable). 🤔

samcunliffe: But if you promise not to change the global, perhaps we're OK?

tests/test_integration/test_frame_extraction.py
```diff
@@ -1,51 +1,490 @@
...
 
 
-class TestsFrameExtraction:
-    def test_extension_case_insensitive(self, input_video_dir):
-        """
-        Tests that the function that computes the list of SLEAP videos
-        is case-insensitive for the user-input extension.
+@pytest.fixture()
+def cli_inputs_dict(tmp_path: Path) -> dict:
```
samcunliffe, line 23: Is this a fixture taking `tmp_path` ... the fixture?!
TIL that this was possible.

tests/test_integration/test_frame_extraction.py
```diff
@@ -1,51 +1,490 @@
...
+
+def check_output_files(list_input_videos: list, cli_dict: dict):
+    """Run assertions on output files from frame extraction
+
+    Parameters
+    ----------
+    list_input_videos : list
+        list of videos used for frame extraction
+    cli_dict : dict
+        a dictionary with the parameters of the frame extraction
```
samcunliffe, line 234: Incidentally (this is a bit bike-sheddy) I don't mind not having a pramaters list for the test functions if they're fixtures. 

It's not like the test functions are public or form part of your API or anything.

... also this file is quite long as it is.

samcunliffe (approved):
All looks good. 🫐 

My comments are mostly questions. I have no substantive comments on the tests themselves.

* You're missing a few `test_flibble() -> None:` typehints. Don't know if they're important in tests but you have some typehints but not all.
* The file is quite long. Is there a smart way to break it up? Things that come to mind:
    - Remove `Parameters` from the docstrings for tests?
    - Make more explicit test function names so you can do away with docstrings altogether?!
    - **cough** longer line length **cough** in the linter.

tests/test_integration/test_frame_extraction.py
sfmig: I have this doubt so many times...

sfmig: but yeah happy to make it a global in this case

sfmig:
thanks for the feedback @samcunliffe!
I reduced the length of the tests file by:
- reducing docstrings in fixtures and tests,
- factoring out the fixtures, and 
- using parametrisation to combine the tests that run the CLI app with and without CLI inputs. 

I think it looks a bit more neat, with the caveat that now I am using the mocked app in both test cases, so in one case (the empty CLI args case) I'm unnecessarily detaching from the real use case. Let me know any thoughts!

tests/test_integration/test_frame_extraction.py
```diff
@@ -1,51 +1,193 @@
...
+    ],
+)
+def test_frame_extraction_one_video(
+    input_video: str,
+    cli_inputs_fixture: str,
+    cli_inputs_dict: dict,
+    mock_extract_frames_app: typer.main.Typer,
+    request: pytest.FixtureRequest,
+) -> None:
+    """Test frame extraction on one video, using default CLI arguments.
```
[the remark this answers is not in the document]
sfmig: oh yeah good point 🦅

PR #26, opened 2023-08-15
Frame extraction in the cluster

sfmig:
A HowTo guide to run a frame extraction job in the SWC cluster


nikk-nikaznan (approved):
Seems very clear to me, thank you for this

Issue #14, opened 2023-08-07
Consider re-enconding videos 


sfmig:
In some of the frame extraction runs, we get the following error:
`KeyError: "Unable to load frame 332416 from MediaVideo(filename='/ceph/zoo/raw/CrabField/ramalhete_2021/camera1/NINJAV_S001_S001_T002.MOV', grayscale=False, bgr=True, dataset='', input_format='')."`

We also saw it with frame `61938` from 
 `filename='/ceph/zoo/raw/CrabField/ramalhete_2021/camera1/NINJAV_S001_S001_T004.MOV'`

From the SLEAP discussions, this seems related to [the video not being reliably seekable](https://github.com/talmolab/sleap/discussions/1095#discussioncomment-4478688).

We need to inspect whether we need to re-encode the videos to make them seekable and avoid these issues

PR #31, opened 2023-08-17
Bash script to reencode videos

sfmig:
Adding a bash script to reencode videos, following [SLEAP guidelines](https://sleap.ai/help.html?highlight=reliably%20seekable#does-my-data-need-to-be-in-a-particular-format)

~~Should fix issue 14 but likely need a bit more research~~ sadly it didn't, still looking into issue #14


bash_scripts/reencode_videos_sleap.sh
```diff
@@ -0,0 +1,79 @@
...
+    FILEPATH=${INPUT_DATA_LIST[${SLURM_ARRAY_TASK_ID}]}
+    filename_no_ext="$(basename "$FILEPATH" | sed 's/\(.*\)\..*/\1/')"
+    echo "Input video: $FILEPATH"
+
+    ffmpeg -y -i "$FILEPATH" \
+    -c:v libx264 \
+    -pix_fmt yuv420p \
+    -preset superfast \
+    -crf 15 \
+    "$OUTPUT_DIR/$OUTPUT_SUBDIR/$filename_no_ext.mp4"
```
samcunliffe, line 63: Do we want a [`-loglevel verbose`, or a `-v`](https://ffmpeg.org/ffmpeg.html#Generic-options)?

And do you `tee` the output (of this script) somewhere?
Actually -- and I'm showing my `slurm`-ignorance -- do you get that anyway in the job logs?

bash_scripts/run_frame_extraction_array.sh
```diff
@@ -45,24 +49,24 @@ fi
...
-PARAM_PER_CLUSTER=8
+PARAM_PER_CLUSTER=10
 
 # ----------------------
 # script location
 # ----------------------
 # assumes repo located at '/ceph/scratch/sminano'
 SCRATCH_PERSONAL_DIR=/ceph/scratch/sminano
-SCRIPT_DIR=$SCRATCH_PERSONAL_DIR/crabs-exploration/"bboxes labelling"
+SCRIPT_DIR=$SCRATCH_PERSONAL_DIR/crabs-exploration/bboxes_labelling
```
samcunliffe, line 69:
```suggestion
SCRIPT_DIR=$SCRATCH_PERSONAL_DIR/crabs-exploration/crabs/bboxes_labelling
```

samcunliffe (approved):
Yeah, I like the flatter directory structure. And the less specific directory names. Or, at least, it's more to my taste.

bash_scripts/reencode_videos_sleap.sh
sfmig: > Do we want a -loglevel verbose, or a -v?

In the "upgrade" of this script, I reencode the video and then extract frames from there (PR #78). For that, I had a go at running `ffmpeg -v`, but it was a bit too verbose.... In fact the default info output level already prints out the parameters of the reencoding, so I ended up only adding a line with `ffmpeg --version` to make sure that was registered. I add now the `ffmpeg --version` line here, but I think this script and the reencode+extract frames one are likely better as one...

And yes, as you say I get`ffmpeg`'s stdout and stderr in slurm's logs (.err and .out files)

PR #72, opened 2023-10-30
Utils for combining annotations

### LINKED ISSUE #70: Combine bboxes labelling util functions in one module?
Consider whether to combine the [bboxes labelling](https://github.com/SainsburyWellcomeCentre/crabs-exploration/tree/main/bboxes_labelling) helper functions into one util module

sfmig:
Made an `annotations_utils` modules with functions to:
- combine several VIA JSON files into one (`combine_multiple_via_jsons`)
- generate a COCO-formatted annotation file from a VIA JSON file (`convert_via_json_to_coco`)

I ended removed the previous `combine_coco` function (which appended COCO files two-by-two) as I found combining VIA JSON files was easier (we just need to append the data, no need to keep track of the ids), and so we could write more easily a function that combines several files in one go.

The previous `convert_coco` function is the basis for the `convert_via_json_to_coco` function with minor additions.

Also included a minor fix to the frame extraction test which were failing before.

Closes #70


bboxes_labelling/combine_and_format_annotations.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+    # Convert to COCO
+    coco_out_fullpath = convert_via_json_to_coco(json_out_fullpath)
+
+    return coco_out_fullpath
+
+
+if __name__ == "__main__":
+    parent_dir_via_jsons = (
```
nikk-nikaznan, line 31: I know the latest annotations are here, but maybe not hardcoded here will be better? We can have it as an args in bash script. What do you think?

tests/test_coco_and_via_conversions.py
```diff
@@ -0,0 +1,342 @@
...
+def via_json_1() -> str:
+    """Return path to a sample VIA JSON file.
+
+    Returns
+    -------
+    str
+        path to a sample VIA JSON file 1.
+    """
+    # Return path to sample VIA (Visual Image Annotator) JSON file 1
+    return str(Path("tests/data/COCO_VIA_JSONS/VIA_JSON_1.json").resolve())
```
nikk-nikaznan, line 25: I think we don't need tests in the path because we are in tests directory?
`return str(Path("data/COCO_VIA_JSONS/VIA_JSON_1.json").resolve())`

tests/test_coco_and_via_conversions.py
```diff
@@ -0,0 +1,342 @@
...
+def via_json_2() -> str:
+    """Return path to a sample VIA JSON file.
+
+    Returns
+    -------
+    str
+        path to a sample VIA JSON file 2.
+    """
+    # Return path to sample VIA JSON file 2
+    return str(Path("tests/data/COCO_VIA_JSONS/VIA_JSON_2.json").resolve())
```
nikk-nikaznan, line 39: Same here

nikk-nikaznan (changes requested):
Looking very good! Good move to create the utils for this. Run this locally including the pytest. Apart from directory, seems fine

tests/test_coco_and_via_conversions.py
sfmig: yes, good catch! I changed it so that it is relative to the path of the test...py file now, so that it works regardless of the directory you are running pytest from

bboxes_labelling/combine_and_format_annotations.py
sfmig: yes, 100%. I was doubting whether to add this to the PR, but I figured this way it's at least under version control...

If it's alright I would say we merge this for now, and open an issue on refactoring / restructuring the whole codebase, with an explicit reference to this?

nikk-nikaznan (approved):

PR #77, opened 2023-11-13
Add exclude pattern to combining annotations

sfmig:
Add option to exclude files from the combination operation.

It allows us to use a regex pattern to exclude annotation files we don't want to be part of the combined annotation file.


crabs/bboxes_labelling/annotations_utils.py
```diff
@@ -79,6 +84,17 @@ def combine_multiple_via_jsons(
...
 
+    # Apply exclude pattern if required
+    if exclude_pattern:
+        list_json_files = [
+            js
+            for js in list_input_json_files
+            if not re.search(exclude_pattern, str(js))
+        ]
+    else:
+        list_json_files = list_input_json_files.copy()
```
samcunliffe, line 95: Any reason to copy the list? I'd probably just filter it in place...
```python
if exclude_pattern:
    not_excluded = lambda js: not re.search(exclude_pattern, str(js))
    list_input_json_files = [js for js in list_input_json_files if not_excluded(js)]
```

With a `lambda` because I'm a hipster.

crabs/bboxes_labelling/combine_and_format_annotations.py
```diff
@@ -14,38 +15,44 @@
...
     ----------
     parent_dir_via_jsons : str
         path to the parent directory containing VIA JSON files
+    exclude_pattern : Optional[str], optional
+        a regex pattern that matches files to exclude.
+        By default, None.  E.g.: "\w+_coco_gen.json$"
     via_default_dir : str
-        The default directory in which to look for images for the VIA project.
-        A full path is required.
+        he default directory in which to look for images for the VIA project.
```
samcunliffe, line 32:
```suggestion
        The default directory in which to look for images for the VIA project.
```

crabs/bboxes_labelling/combine_and_format_annotations.py
```diff
@@ -14,38 +15,44 @@
...
     """
 
-    # Get list of VIA JSON files
-    list_json_files = [
+    # Get list of all JSON files in directory
+    list_input_json_files = [
         x
         for x in Path(parent_dir_via_jsons).glob("*")
         if x.is_file() and str(x).endswith(".json")
     ]
```
samcunliffe, line 50:
```suggestion
    all_files = Path(parent_dir_via_jsons).glob("*")
    is_a_json = lambda x: x.is_file() and str(x).endswith(".json")
    list_input_json_files = [x for x in all_files if is_a_json(x)]
```

Don't know if you agree but with a nicely named filter function and a nicely named variable, you can probably do away with the comment and get it all on one line.

samcunliffe (commented):
Looks good, it's possible I just can't get it to work because I'm being dim.

crabs/bboxes_labelling/annotations_utils.py
sfmig: wooo fancy! ✨ me gusta

sfmig: but [Ruff disapproves](https://docs.astral.sh/ruff/rules/lambda-assignment/) 😢

sfmig: removed the copy but kept list comprehension

crabs/bboxes_labelling/combine_and_format_annotations.py
sfmig: I like it, reads more easily than the first version. But I kept the conditional in the list comprehension because [ruff disapproves lambda assignments ](https://docs.astral.sh/ruff/rules/lambda-assignment/).

sfmig:
thanks @samcunliffe for having a look!

Re the fake jsons test, I reproduced exactly your commands and I got: `ValueError: Error decoding JSON data from file: test_fake_jsons/bad-1.json` - which makes sense since there is nothing in the json file right? We are also expecting json files in the VIA format.

If you are wondering why the `list_input_json_files` includes the "bad" ones, I would still expect that because the filtering occurs when you combine them (rather than when you read the CLI arguments).

This test in which we copy a sample annotation file should work:
```
mkdir test_fake_jsons
for i in {1..10}; do cp tests/data/COCO_VIA_JSONS/VIA_JSON_1.json test_fake_jsons/good-$i.json; done
for i in {1..10}; do cp tests/data/COCO_VIA_JSONS/VIA_JSON_1.json test_fake_jsons/bad-$i.json; done
combine-annotations test_fake_jsons --exclude-pattern "bad-*.json"
```

crabs/bboxes_labelling/annotations_utils.py
samcunliffe: Sigh. I love named lambdas, they're great, especially for readability.

samcunliffe:
> If you are wondering why the list_input_json_files includes the "bad" ones, I would still expect that because the filtering occurs when you combine them (rather than when you read the CLI arguments).

Ah right. I assumed you'd filter the files _before_ you opened them.

samcunliffe (approved):

sfmig:
> Ah right. I assumed you'd filter the files before you opened them.

@samcunliffe I do, the error message just shows what you passed as a CLI argument right?

samcunliffe:
> > Ah right. I assumed you'd filter the files before you opened them.
> 
> @samcunliffe I do, the error message just shows what you passed as a CLI argument right?

Ahhh sorry. I'm being slow. So the error is thrown after filtering, before opening the files ✅. But the fact that it threw an error **_at all_** makes typer dump the CLI arguments for your information.

samcunliffe:
Confirmed that it works with toy Jsons containing actual content. 👍

PR #78, opened 2023-11-13
Frame extraction with reencoding

sfmig:
Some improvements / additions to the frame extraction slurm script
- optionally reencode videos before extracting frames
- save SLURM logs along with the extracted frames
- save SLURM logs along with the reencoded videos 
- derive the file extension from the input video

I deleted the frame extraction only script and the script to extract frames locally.

This PR is a first step towards a more general bash script that would make use of the CLI tools and that we wouldn't edit for each run - some thoughts on that in this draft PR #88 (see [this file specifically](https://github.com/SainsburyWellcomeCentre/crabs-exploration/pull/88/files#diff-95c9695dd1abff98368e08e4552e4bf6cb1a09c2a50d73e684d3a629a1f4b0fd))


bash_scripts/run_reencode_and_frame_extraction_array.sh
```diff
@@ -0,0 +1,166 @@
...
+            /$REENCODED_VIDEOS_SUBDIR/"$filename_no_ext"_RE.slurm_array.$SLURM_ARRAY_JOB_ID-$SLURM_ARRAY_TASK_ID.$ext
+        done
+    fi
+
+    # Frame extraction logs
+    # Move logs for this job to subdir with extracted frames
+    for ext in err out
+    do
+        mv slurm_array.$SLURMD_NODENAME.$SLURM_ARRAY_JOB_ID-$SLURM_ARRAY_TASK_ID.$ext /$LOG_DIR
+    done
```
samcunliffe, line 163:
```suggestion
    mv slurm_array.$SLURMD_NODENAME.$SLURM_ARRAY_JOB_ID-$SLURM_ARRAY_TASK_ID.{err,out} /$LOG_DIR
```
.. no? (I think you need the `ext` variable in the loop above though.)

samcunliffe: p.s. please test. It's been ages since I wrote bash scripts for a cluster.

crabs/bboxes_labelling/extract_frames_to_label_w_sleap.py
```diff
@@ -325,12 +325,10 @@ def extract_frames_to_label_from_video(
...
 
             # If read successfully: save to file
-            # file naming format: parentdir_videoname_frame_XXX.png
+            # file naming format: videoname_frame_XXX.png
             else:
                 file_path = video_output_dir / Path(
-                    f"{Path(vid_str).parent.stem}_"
-                    f"{Path(vid_str).stem}_"
-                    f"frame_{frame_idx:06d}.png",
+                    f"{Path(vid_str).stem}_" f"frame_{frame_idx:06d}.png",
```
samcunliffe, line 331: How scalable is the 6 digits in frame number? Worth a couple more leading zeroes in there for safety? Or not needed.

samcunliffe (approved):
It's been a while since I've written a longish bash script.

Even though it's a bit out of scope for this PR, we should think about how to test these guys. Like could we call them in CI with a `--dryrun` flag or something?

You might want to ignore my comments 👇  if there are plans in #88 (I didn't dig into that yet because: rebase.)

crabs/bboxes_labelling/extract_frames_to_label_w_sleap.py
sfmig: for now all videos have less than 999999 frames (max is 345048 frames), but will had a couple of leading 0s for safety 🧷

bash_scripts/run_reencode_and_frame_extraction_array.sh
sfmig: yeah it seems `cp` doesn't admit parameter expansion in the same way as `mv`
see [here](https://stackoverflow.com/a/5580894)

sfmig:
> It's been a while since I've written a longish bash script.
> 
> Even though it's a bit out of scope for this PR, we should think about how to test these guys. Like could we call them in CI with a `--dryrun` flag or something?
> 
> You might want to ignore my comments 👇 if there are plans in #88 (I didn't dig into that yet because: rebase.)

Yes, testing this would be ideal. I had a quick look and [bats](https://github.com/bats-core/bats-core) seems like a very nice tool for this. IIUC, [it can be run in CI ](https://github.com/bats-core/bats-core/blob/master/.github/workflows/tests.yml) too. 

Right now we manually edit this script for the different frame extraction jobs. So testing this would be cumbersome. My idea with #88 is to have a bash script that we keep intact, and to which we pass any required parameters as command line arguments. Specifically, we would pass a config.json file that we would read and print to the logs. This way we keep track of what input parameters we passed but the bash script is not edited every time.

So for now I would suggest to merge this PR and add these comments to PR #88. There are other options for unit testing bash scripts, I am adding a brief overview to PR #88 too.

samcunliffe:
> So for now I would suggest to merge this PR...

Agreed.

PR #81, opened 2023-11-16
How to create GIN private repo for crab data

sfmig:
A guide on how to set up a GIN private repo for the crab data. 

To be shared with NIU team when we finalise it.

I added my current thoughts about how it works. Any feedback on making things clearer / less verbose is more than welcome!

PR #75, opened 2023-11-01
Update extract frames guide

sfmig:
- update paths in extract frames guide
- some small edits in labelling guide (I got distracted...)
- remove dev version of labelling guide

PR #86, opened 2023-11-17
Light refactoring

sfmig:
A first chunk of light refactoring:
- an example of avoiding [long parameter list](https://refactoring.guru/smells/long-parameter-list)
- reorganised the directories (moved bash scripts, guides and notebooks to the root level)

I figured we could merge to main this new structure and then I continue with closing currently open PRs. 

Next steps towards that:
- add stereo-calibration work (as it is in current state) 
    - PR #2
- add bash scripts:
     - to run video reencoding (PR #31) 
     - to run reencoding+frame extraction (PR #78)
- add/update guides
    - PRs #81, 
    - PR #75
- add exclude pattern feature to combine annotations, with tests 
    - PR #77
- finish test for frame extraction 
    - PR #24


samcunliffe (approved):
👍

crabs/dense_optical _flow/estimate_optical_flow_on_video.py
samcunliffe, line 1: Wait. We've removed this but didn't re-add. Presumably `crabs/dense_optical_flow/*.py`.

samcunliffe (changes requested):

crabs/dense_optical _flow/estimate_optical_flow_on_video.py
sfmig: Yeah that is actually intended! We are not using this optic flow estimation anymore (we are doing a simpler computation for it instead). I think it's good enough to have it in the history if we ever need to retrieve it, but we can keep the codebase cleaner by removing it from the src for now.

Let me know if all ok

samcunliffe (approved):

PR #103, opened 2024-02-21
Nikna/pl dev

nikk-nikaznan:
Seems quite big, but it is mostly because I moved some functions around.
This branch is implementation of PyTorch-lightning.

- created `datamodule` file where most of the data loader related code is here for both train and evaluation. This is related to the previous PR with some data loader codes seem redundant in training and evaluation.
- some of the functions in datamodule were from `detection_utils`. Some of the code from `evaluation` is now in `detection_utils` as it will be used by inference as well (not yet).
- moving `myFasterRCNNDataset` to it's own file

The PyTorch-lightning implementation can be found in 

- datamodule
- evaluate_model
- models
- train_model

The cool things about data module, easier MLFlow and tensor board implementation.

### DIFF
--- a/crabs/detection_tracking/config/faster_rcnn.yaml
+++ b/crabs/detection_tracking/config/faster_rcnn.yaml
@@ -1,7 +1,7 @@
 num_epochs: 1
 learning_rate: 0.00005
 wdecay: 0.00005
-batch_size: 4
+batch_size: 2
 num_classes: 2
 image_stack: False
 Sort: False
--- a/crabs/detection_tracking/config/faster_rcnn.yaml
+++ b/crabs/detection_tracking/config/faster_rcnn.yaml
@@ -1,4 +1,4 @@
-num_epochs: 250
+num_epochs: 1
 learning_rate: 0.00005
 wdecay: 0.00005
 batch_size: 4
--- a/crabs/detection_tracking/config/faster_rcnn.yaml
+++ b/crabs/detection_tracking/config/faster_rcnn.yaml
@@ -1,7 +1,7 @@
 num_epochs: 250
 learning_rate: 0.00005
 wdecay: 0.00005
-batch_size: 8
+batch_size: 4
 num_classes: 2
 image_stack: False
 Sort: False
--- a/crabs/detection_tracking/datamodule.py
+++ b/crabs/detection_tracking/datamodule.py
@@ -1,0 +1,241 @@
+import json
+
+import lightning as pl
+from sklearn.model_selection import train_test_split
+from torch.utils.data import DataLoader
+from torchvision.transforms import transforms
+
+from crabs.detection_tracking.myfaster_rcnn_dataset import myFasterRCNNDataset
+
+
+def collate_fn(batch):
+    """
+    Collates a batch of samples into a structured format that is expected in Pytorch.
+
+    This function takes a list of samples, where each sample can be any data structure.
+    It filters out any `None` values from the batch and then groups the
+    remaining samples into a structured format. The structured format
+    is a tuple of lists, where each list contains the elements from the
+    corresponding position in the samples.
+
+    Parameters
+    ----------
+    batch : list
+        A list of samples, where each sample can be any data structure.
+
+    Returns
+    -------
+    collated : Optional[Tuple[List[Any]]]
+        A tuple of lists, where each list contains the elements from the corresponding
+        position in the samples. If the input batch is empty or contains only `None`
+        values, the function returns `None`.
+    """
+    batch = [sample for sample in batch if sample is not None]
+
+    if len(batch) == 0:
+        return None
+
+    return tuple(zip(*batch))
+
+
+def get_train_transform() -> transforms.Compose:
+    """
+    Get a composed transformation for training data for data augmentation and
+    transform the data to tensor.
+
+    Returns
+    -------
+    transforms.Compose
+        A composed transformation that includes the specified operations.
+
+    Notes
+    -----
+    This function returns a composed transformation for use with training data.
+    The following transforms are applied in sequence:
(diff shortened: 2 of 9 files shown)


crabs/detection_tracking/datamodule.py
```diff
@@ -0,0 +1,241 @@
...
+        super().__init__()
+        self.main_dir = main_dir
+        self.annotation = annotation
+        self.batch_size = batch_size
+        self.seed_n = seed_n
+
+    def prepare_data(self):
+        pass
+
+    def setup(self, stage=None):
```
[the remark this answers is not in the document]
nikk-nikaznan: `prepare_data` is usually for downloading the data. This is like the convention class for PyTorch_lightning data module. I think technically, we can delete it as we not using it.

crabs/detection_tracking/datamodule.py
```diff
@@ -0,0 +1,241 @@
...
+
+    Attributes:
+    -----------
+    main_dir : str
+        The main directory path containing the dataset.
+    annotation : str
+        The filename of the COCO annotation JSON file.
+    batch_size : int
+        Batch size for data loading.
+    seed_n : int
```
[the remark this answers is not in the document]
nikk-nikaznan: The reason for having the seed here is to make sure we not using the same frames in training and evaluation. The current code has separate training and evaluation. So we run training, save the model. Then when we want to see the result, we run the evaluation, and load the model. so by setting the seed, we can control the split properly. I will in my plan to merge them into one code. so when we run the training, it will save the model and do the evaluation as well. But during development, it is always better to have them separately in my opinion.
The seed_n is a part of args now.

crabs/detection_tracking/datamodule.py
```diff
@@ -0,0 +1,241 @@
...
+            batch_size=self.batch_size,
+            shuffle=True,
+            num_workers=4,
+            collate_fn=collate_fn,
+            persistent_workers=True,
+        )
+
+    def val_dataloader(self):
+        """
+        Returns the data loader for the evaluation set.
```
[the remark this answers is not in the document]
nikk-nikaznan: During the training we want to shuffle the data, so that for every epoch the order of the data will be different so the network not going to cheat by any means having the data in order. Usually we don't shuffle during test so the results are consistent when we compared against different runs.
The same with persistent_workers, we need it to be true when dealing with big data. During test as we are dealing with smaller batch number, we don't really need it.

crabs/detection_tracking/datamodule.py
```diff
@@ -0,0 +1,241 @@
...
+                "09.08.2023-03-Left",
+                "10.08.2023-01-Left",
+                "10.08.2023-01-Right",
+            ]
+            all_ids = []
+
+            for image_info in self.coco_data["images"]:
+                image_id = image_info["id"]
+                image_file = image_info["file_name"]
+                video_file = image_file.split("_")[1]
```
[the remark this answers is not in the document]
nikk-nikaznan: It works okay for now, as we not excluding any video from Sep2023 dataset. But yes need to revisit this

README.md
```diff
@@ -5,3 +5,29 @@
...
+For training
+
+```bash
+python crabs/detection_tracking/train_model.py --main_dir {parent_directory_of_frames_and_annotation} --annotation_file {annotation_file.json}
+```
+
+For evaluation
+
+```bash
+python crabs/detection_tracking/train_model.py --model_dir {directory_to_saved_model} --main_dir {parent_directory_of_frames_and_annotation} --annotation_file {annotation_file.json}
```
sfmig, line 32:
```suggestion
python crabs/detection_tracking/evaluate_model.py --model_dir {directory_to_saved_model} --main_dir {parent_directory_of_frames_and_annotation} --annotation_file {annotation_file.json}
```

sfmig: thanks for adding this! ✨

crabs/detection_tracking/config/faster_rcnn.yaml
sfmig, line 1: looks great

crabs/detection_tracking/datamodule.py
```diff
@@ -0,0 +1,254 @@
...
+    Returns
+    -------
+    collated : Optional[Tuple[List[Any]]]
+        A tuple of lists, where each list contains the elements from the corresponding
+        position in the input batch of samples. If the input batch is empty or contains only
+        `None` values, the function returns empty tuple.
+
+    Example
+    -------
+    >>> collate_fn([[1, 2, 3], [4, 5, 6], None, [7, 8, 9]])
```
sfmig, line 36: the example is very useful!

crabs/detection_tracking/datamodule.py
```diff
@@ -0,0 +1,254 @@
...
+
+    def prepare_data(self):
+        """
+        Function to prepare the data (i.e., data downloading)
+        """
+        pass
+
+    def setup(self, stage=None):
+        """
+        Sets up the data loader for the specified stage
```
sfmig, line 139: I think "sets up the dataloader" is still a bit ambiguous, and doesn't add much than the function name...
We could maybe say more specifically: "computes train and test IDs for the dataloader", or  "computes the train/test splits".

I was a bit confused about stage None so did a bit of googling....apparently [I'm not the only one](https://stackoverflow.com/a/73261116) 😅 . It seems it's an historic thing:
> The type is annotated optional and the default value is None for historical reasons. The values will always be one of "fit", "validate", "test", "predict".

Maybe we could remove it?

sfmig (approved):
Looking good! A couple of comments but should be good to go. 

There is a few (2?) unresolved comments from the previous review (not sure if you missed them) but they aren't vital IMO, so feel free to leave as is.

PR #84, opened 2023-11-16
Entrypoints.

samcunliffe:
## Continues

* #73

PR #107, opened 2024-02-26
Entrypoints for frame extraction and for combining annotations

sfmig:
Includes basic smoketests.

It's a copy of PR #84, which we started during a pair-programming session. IIRC, I was typing (the branch has my initials?) so I think it makes more sense if you review - but do let me know if you think otherwise. 

For now I will close the other one to avoid confusion


samcunliffe (approved):
>  I think it makes more sense if you review

🫐

PR #114, opened 2024-03-07
cleaned up mlflow and add readme

nikk-nikaznan:
- add readme on how to run mlflow
- add experiment name 
- add log hyper parameters

### DIFF
--- a/crabs/detection_tracking/train_model.py
+++ b/crabs/detection_tracking/train_model.py
@@ -1,4 +1,5 @@
 import argparse
+import datetime
 
 import lightning as pl
 import torch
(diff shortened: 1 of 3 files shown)


crabs/detection_tracking/train_model.py
```diff
@@ -52,11 +53,19 @@ def train_model(self):
             self.main_dirs, annotations, self.config, self.seed_n
         )
 
-        lightning_model = FasterRCNN(self.config)
+        timestamp = datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+        run_name = f"experiment_{timestamp}"
```
sfmig, line 57:
```suggestion
        run_name = f"run_{timestamp}"
```
I find it confusing to call the run `experiment_`, since in MLflow an experiment has multiple runs. Maybe simply `run_`?

crabs/detection_tracking/train_model.py
```diff
@@ -111,6 +120,12 @@ def main(args) -> None:
         default="gpu",
         help="accelerator for pytorch lightning",
     )
+    parser.add_argument(
+        "--experiment_name",
+        type=str,
+        default="Sept2023",
+        help="the run name for mlflow for example the dataset used for the run",
```
sfmig, line 127:
```suggestion
        help="the name for the experiment in MLflow, under which the current run will be logged. For example, the name of the dataset could be used, to group runs using the same data.",
```
Here the MLflow concepts for "run" and an "experiment" are a bit mixed up here, but this should make it clearer.

sfmig (approved):
MLflow looks very cool 🌟
Small comments to avoid confusion between `run` and `experiment` concepts in MLflow. I also added the mlflow dependency.

I couldn't get a GPU node but hopefully I can try later today.

PR #120, opened 2024-03-14
Make CLI arguments less verbose

sfmig:


sfmig:
closing - it was a split of a larger PR that I ended up splitting in another way

PR #121, opened 2024-03-15
Dataloader: tests for datasets module (3/3)

sfmig:
Started a module for dataset tests, for now including the 'exclude files' tests.

Small fixes and refactoring of the dataset module (the tests successfully pointed to a case I had missed 👏 ).

Moved `conftest.py` to the root directory since "[Requiring plugins using `pytest_plugins` variable in non-root conftest.py files is deprecated](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#requiring-plugins-in-non-root-conftests)".


nikk-nikaznan (approved):
Apart from getting DeprecationWarning (probably my env), everything else looking good!

PR #122, opened 2024-03-15
Dataloader: tests for train script CLI arguments (2/3)

sfmig:
Started a module for the training script tests, for now just including tests for the CLI arguments.


nikk-nikaznan (approved):
Looking good!  😊

PR #127, opened 2024-03-18
An alternative way to exclude files

sfmig:
It doesn't transmit the coco object of the dataset or the list of image files though, so not convinced.

Saving the annotation file with the excluded files removes feels a bit overkill but it's more reproducible.

rebase after dataloader #123


samcunliffe:
Closing for now, for housekeeping reasons. Can revisit and reopen later if still needed.

PR #135, opened 2024-03-26
pytest for detector

nikk-nikaznan:
Some test for

- `detection_utils`
- `datmodules`
- `evaluate`
- `visualisation`


tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,107 @@
...
+    draw_bbox,
+    draw_detection,
+    save_images_with_boxes,
+)
+
+
+def test_draw_bbox():
+    image = np.zeros((100, 100, 3), dtype=np.uint8)
+    draw_bbox(image, (10, 10), (20, 20), (0, 255, 0))
+    cv2.imwrite("bbox_test.png", image)
```
sfmig, line 19: It is a bit odd that the image is saved to the current working directory - seems like a good idea to use [pytest's `tmp_path` fixture](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html). It creates a temporary directory unique to the test invocation, and pytest handles cleanup for us after testing.

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+@pytest.fixture
+def trained_model():
+    return MagicMock()
+
+
+@pytest.fixture
+def score_threshold():
+    return 0.5
```
sfmig, line 81: do we need a fixture for this? It seems we can just set `score_threshold=0.5` in the test below. It would maybe be a good idea to parametrise the test for different `score_threshold` values.

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,107 @@
...
+    image_with_boxes = draw_detection(imgs, annotations, None)
+    assert image_with_boxes is not None
+
+
+@pytest.fixture
+def test_dataloader():
+    return MagicMock()
+
+
+@pytest.fixture
```
sfmig, line 74: I usually put all the fixtures at the top but I don't think there is a strong convention - maybe worth a search

sfmig: I would actually suggest removing these fixtures - see comment below.

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,107 @@
...
+
+    save_images_with_boxes(
+        test_dataloader,
+        trained_model,
+        score_threshold,
+    )
+
+    assert mock_makedirs.called_once_with("results", exist_ok=True)
+    assert mock_draw_detection.call_count == len(test_dataloader)
+    assert mock_imwrite.call_count == len(test_dataloader)
```
sfmig, line 107: do we need `MagicMock` for the detections? It feels like it could be a variable.

IIUC, the main thing is to mock the call to the `trained_model` so that it returns the detections we predefine. I feel like this test could be further simplified by:
- Removing the fixtures for `trained_model` and `test_dataloader` (and just have them inside the test).
- Removing the `draw_detection` mock. I would just keep the `cv2.imwrite` call. This is mainly because  the calls to `cv2.imwrite` follow the calls to `draw_detection` and it feels a bit redundant to me to check both are called.

Maybe `test_dataloader` could also be just a small dataloader for testing too.

```python
@patch("cv2.imwrite")
@patch("os.makedirs")
def test_save_images_with_boxes(
    mock_makedirs,
    mock_imwrite,
):
    detections = [
        {
            "scores": torch.tensor([0.9]),
            "labels": torch.tensor([1]),
            "boxes": torch.tensor([[30, 30, 40, 40]]),
        },
        {
            "scores": torch.tensor([0.8]),
            "labels": torch.tensor([2]),
            "boxes": torch.tensor([[70, 70, 80, 80]]),
        },
    ]
    trained_model = MagicMock()
    test_dataloader = MagicMock()
    trained_model.return_value = detections

    save_images_with_boxes(
        test_dataloader,
        trained_model,
        score_threshold=0.5,
    )

    assert mock_makedirs.called_once_with("results", exist_ok=True)
    assert mock_imwrite.call_count == len(test_dataloader)
```

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,107 @@
...
+    annotations = [{"boxes": torch.tensor([[10, 10, 20, 20]])}]
+    detections = [
+        {
+            "scores": torch.tensor([0.9]),
+            "labels": torch.tensor([1]),
+            "boxes": torch.tensor([[30, 30, 40, 40]]),
+        }
+    ]
+    image_with_boxes = draw_detection(imgs, annotations, detections)
+    assert image_with_boxes is not None
```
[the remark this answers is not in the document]
nikk-nikaznan: Yes the test wasn't the best, but I think it is useful to see how the function handle the arguments, considering some of it is optional?

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,107 @@
...
+import torch
+
+from crabs.detection_tracking.visualization import (
+    draw_bbox,
+    draw_detection,
+    save_images_with_boxes,
+)
+
+
+def test_draw_bbox():
```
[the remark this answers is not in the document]
nikk-nikaznan: I have added different tests to just see if there is any drawing happened

nikk-nikaznan:
Thank you Sofia for the help. I have made some changes, let me know what you think.

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,123 @@
...
+            "labels": torch.tensor([1]),
+            "boxes": torch.tensor([[30, 30, 40, 40]]),
+        },
+        {
+            "scores": torch.tensor([0.8]),
+            "labels": torch.tensor([2]),
+            "boxes": torch.tensor([[70, 70, 80, 80]]),
+        },
+    ]
+    trained_model = MagicMock()
```
sfmig, line 112: these mock utils look very cool, I didn't know you could do this 🤓

samcunliffe (commented):
🫐 

This all looks good. And my comments are really just polishing. I tend to agree with @sfmig that a _slightly_ stupid regression test is probably better than not testing.

sfmig (changes requested):
just requesting changes cause I think the parametrization of `test_draw_detection` is still missing 😬

tests/test_unit/test_visualization.py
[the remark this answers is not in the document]
sfmig: I think @nikk-nikaznan came up with a simpler case that may work for this - basically a simulated crab images that is all black except the box. We can "manually" build that with numpy and compare the output of `draw_bbox` to that.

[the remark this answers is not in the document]
samcunliffe: > a simulated crab image that is all black except the box
▪️ 

That's totally fine, and don't change now. But having a small "real data" image might be useful in the future for other tests.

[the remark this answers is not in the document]
nikk-nikaznan: Yes agree. I am making sample data now, small "real data" maybe 4 of them with some annotations? I will try to resize it to smaller size. Will be useful to test the model related code as well.

tests/test_unit/test_detection_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    expected_annot_paths = []
+    for dataset_dir, annot_file in zip(input_datasets, input_annotation_files):
+        expected_annot_paths.append(
+            str(Path(dataset_dir) / "annotations" / annot_file)
+        )
+
+    assert (
+        prep_annotation_files(input_annotation_files, input_datasets)
+        == expected_annot_paths
+    )
```
sfmig, line 70: I think we lost one test along the way:

```python
@pytest.mark.parametrize(
    "input_datasets, input_annotation_files",
    [
        ([DATASET_1], ["path/to/file1.json"]),
        ([DATASET_1, DATASET_2], ["path/to/file1.json", "path/to/file2.json"]),
    ],
)
def test_prep_annotation_files_with_fullpaths(
    input_datasets, input_annotation_files
):
    expected_annot_paths = []
    for annot_file in input_annotation_files:
        expected_annot_paths.append(annot_file)

    assert (
        prep_annotation_files(input_annotation_files, input_datasets)
        == expected_annot_paths
    )
```

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@pytest.mark.parametrize(
+    "top_left, bottom_right, color, label_text",
+    [
+        ((10, 10), (50, 50), (0, 255, 0), None),
+        ((10, 10), (50, 50), (0, 255, 0), "Test Label"),
+    ],
+)
+def test_draw_bbox(sample_image, top_left, bottom_right, color, label_text):
```
sfmig, line 27: this test seems to test slightly the wrong thing:
- If a text is passed, it checks the text was drawn (but it doesn't check the box was)
- If no text is passed, then it checks the box was drawn.

I think I would either:
- Option 1: modify this test so that the box is always checked (regardless of whether a text is passed or not).
- Option 2: keep only the text test (I think it was like this in the previous review). For the box, you already have the `test_draw_bbox_crop`, which is stricter than the check you do here. 

I think Option 2 is best. For the text test, this is what you had before:

```python
def test_draw_bbox_with_label(sample_image):
    top_left = (10, 10)
    bottom_right = (50, 50)
    color = (0, 255, 0)
    label_text = "Test Label"

    draw_bbox(sample_image, top_left, bottom_right, color, label_text)

    # Check if label text is drawn
    assert np.any(sample_image[10:20, 10:100] == color)
```

Small improvement: can we link the coordinates where we expect the text to be (last line in the snippet above) to the bounding box coordinates?

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,172 @@
...
+                top_left[1] : bottom_right[1], top_left[0] : bottom_right[0]
+            ]
+            == color
+        )  # Check if bounding box is drawn
+
+
+@pytest.mark.parametrize(
+    "top_left, bottom_right, color",
+    [
+        ((10, 10), (15, 15), (0, 255, 0)),
```
sfmig, line 46: Since we are only checking one case I don't think this needs parametrisation for now. 

Or we can add another case (but note it will only work for green boxes).

We can expand in a future PR tho, happy to leave as is!

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,172 @@
...
+        (
+            [
+                {"boxes": torch.tensor([[10, 10, 20, 20]])},
+                {"boxes": torch.tensor([[50, 50, 60, 60]])},
+            ],
+            None,
+        ),
+    ],
+)
+def test_draw_detection_parametrized(annotations, detections):
```
sfmig, line 133:
```suggestion
def test_draw_detection(annotations, detections):
```

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,172 @@
...
+    imgs = [torch.rand(3, 100, 100)]
+    image_with_boxes = draw_detection(imgs, annotations, detections)
+    assert image_with_boxes is not None
+
+
+@pytest.mark.parametrize(
```
sfmig, line 139: same as before, for one case we don't really need to parametrise.

But maybe easier to expand to other cases later? Up to you I think

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,172 @@
...
+        )  # Check if bounding box is drawn
+
+
+@pytest.mark.parametrize(
+    "top_left, bottom_right, color",
+    [
+        ((10, 10), (15, 15), (0, 255, 0)),
+    ],
+)
+def test_draw_bbox_crop(sample_image, top_left, bottom_right, color):
```
sfmig, line 49:
```suggestion
def test_draw_bbox_green(sample_image, top_left, bottom_right, color):
```
maybe?

sfmig (approved):
looking good! 

Some small comments and a larger one on `test_draw_bbox` (I think some checks are missed as it now but it has an easy fix).

I think we also missed a test along the way for the annotations prep.

tests/test_unit/test_visualization.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+@pytest.mark.parametrize(
+    "top_left, bottom_right, color, label_text",
+    [
+        ((10, 10), (50, 50), (0, 255, 0), "Test Label"),
+    ],
+)
+def test_draw_bbox_with_label(
+    sample_image, top_left, bottom_right, color, label_text
```
sfmig, line 44: I think you forgot to pass `label_text` to `draw_bbox`

PR #140, opened 2024-03-27
Fix transforms passed to each dataset split

sfmig:
We were passing the transforms of the training set to all other sets.

With this fix we pass the test transforms to the validation set and the test set.

Feels a bit overkill to compute all three splits every time, but to use `random_split` we need to pass a dataset with transforms already linked to it. Open to suggestions for a nicer way 👀 .

Also: how could we make a test for this?

### DIFF
--- a/crabs/detection_tracking/datamodules.py
+++ b/crabs/detection_tracking/datamodules.py
@@ -89,8 +89,11 @@ class CrabsDataModule(LightningDataModule):
 
     def _compute_splits(
         self,
+        transforms_all_splits: torchvision.transforms,
     ) -> tuple[CrabsCocoDetection, CrabsCocoDetection, CrabsCocoDetection]:
         """Compute train/test/validation splits.
+
+        NOTE: The same transforms are passed to all splits.
 
         The split is reproducible if the same seed is passed.
 
@@ -117,7 +120,7 @@ class CrabsDataModule(LightningDataModule):
         full_dataset = CrabsCocoDetection(
             self.list_img_dirs,
             self.list_annotation_files,
-            transforms=self.train_transform,
+            transforms=transforms_all_splits,  # self.train_transform,
             list_exclude_files=self.config.get(
                 "exclude_video_file_list"
             ),  # get value only if key exists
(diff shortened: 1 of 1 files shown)


crabs/detection_tracking/datamodules.py
```diff
@@ -163,18 +166,27 @@ def setup(self, stage: str):
...
         self.train_transform = self._get_train_transform()
         self.test_transform = self._get_test_val_transform()
         self.val_transform = self._get_test_val_transform()
 
-        # Assign datasets for dataloader depending on stage
+        # Assign datasets for dataloader depending on stage.
         # omitting "predict" stage for now
-        train_dataset, test_dataset, val_dataset = self._compute_splits()
         if stage == "fit":
+            # compute splits passing the relevant transform
```
nikk-nikaznan, line 177: we don't need this condition actually. Lightning will deal with it.
after assigning the transform..
```
self.train_dataset, _, _ = self._compute_splits(self.train_transform)
_, _, self.val_dataset = self._compute_splits(self.val_transform)
_, self.test_dataset, _ = self._compute_splits(self.test_transform)
```

crabs/detection_tracking/datamodules.py
```diff
@@ -117,7 +120,7 @@ def _compute_splits(
         full_dataset = CrabsCocoDetection(
             self.list_img_dirs,
             self.list_annotation_files,
-            transforms=self.train_transform,
+            transforms=transforms_all_splits,  # self.train_transform,
```
nikk-nikaznan, line 123: I overlooked this, so before this we apply train_transform to all the images?

nikk-nikaznan (approved):
I have tested on my end to remove the condition for stage and it works the same. From lighting latest documentation we don't really need that.

In term of the transform, I guess it's fine for now. I will try to dig more better to do this, but it's not making the loading much slower or anything

crabs/detection_tracking/datamodules.py
sfmig: I couldn't find this update in the docs - would you mind linking to it? 

So without the stage we now compute all splits, regardless of the stage. Is that correct?

PR #141, opened 2024-03-27
Data augmentation

sfmig:
Rebase after #203 is merged

---
This PR adds a few data augmentation transforms that we think could be helpful.

Specifically,
- adds some preselected transforms with reasonable values to the config yaml file,
- adds a CLI option `--no_data_augmentation` to skip all data augmentation during training,
- adds a CLI option `--log_data_augmentation` to log the data augmentations linked to the datamodule as MLflow artefacts,
- adds a notebook for visualisation,
- adds data augmentation tests.


nikk-nikaznan (approved):
Cool stuff! I am happy with this, run well and I tried to commented out some variable as well. The only thing, I guess we need to add some guide or in the readme for this. As we never test this properly, so not necessarily, the default one in the config will give the best results. Some might harm the model more. But this will be cool for anyone to start doing ablation study on data augmentation.

sfmig:
aah good call! 
Yeah my idea was to run a study first, then get a rough estimate of what is the best performing set of parameters (rough estimate because I am not optimising the parameters per transform for example. Then have those as a default as you say.

But I think having a guide on how to run a study like this could be helpful - I  opened an [issue](https://github.com/SainsburyWellcomeCentre/crabs-exploration/issues/204).

thanks Nik!

PR #143, opened 2024-04-04
Recover stage argument in setup for datamodule

sfmig:
We need to keep the `stage` argument in the setup of the datamodule, as it seems to be used internally. Otherwise we get:

```
TypeError: CrabsDataModule.setup() got an unexpected keyword argument 'stage'
```

Since it doesn't have a default value, in the eval module we also need to specify it when we call `setup` directly.

### DIFF
--- a/crabs/detection_tracking/datamodules.py
+++ b/crabs/detection_tracking/datamodules.py
@@ -152,7 +152,7 @@ class CrabsDataModule(LightningDataModule):
         """
         pass
 
-    def setup(self, stage: str):  # type: ignore
+    def setup(self, stage: str):
         """Setup the data for training, testing and validation.
 
         Define the transforms for each split of the data and compute them.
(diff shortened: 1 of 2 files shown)


crabs/detection_tracking/datamodules.py
```diff
@@ -152,19 +152,19 @@ def prepare_data(self):
         """
         pass
 
-    def setup(self):
+    def setup(self, stage: str):
```
nikk-nikaznan, line 155: Sorry, I was overlooking this. Yes we need stage argument, we just don't need the condition check, sorry. I set to None or we can do like this both working.

nikk-nikaznan (approved):

crabs/detection_tracking/datamodules.py
sfmig: no worries, I also missed it 🙌

PR #144, opened 2024-04-05
Changes the log to log every epoch

nikk-nikaznan:
The `.log` function cannot force the api to log for every epoch even when set `on_epoch=True` and `on_step=False`.
[Seems like unresolved issue](https://github.com/Lightning-AI/pytorch-lightning/issues/3228).

Instead we do `.log_metric` and set `step=self.current_epoch`. The x-axis will say step but it is in epoch.

Few other small changes:

- log model_filename in the mlflow
- options for fast debugging:

1. `fast_dev_run` only run for one batch and one epoch.
2. `limit_train_batches` run every epochs with limited percentage of training data.

### DIFF
--- a/crabs/detection_tracking/train_model.py
+++ b/crabs/detection_tracking/train_model.py
@@ -47,6 +47,8 @@ class DectectorTrain:
         self.accelerator = args.accelerator
         self.seed_n = args.seed_n
         self.experiment_name = args.experiment_name
+        self.fast_dev_run = args.fast_dev_run
+        self.limit_train_batches = args.limit_train_batches
         self.load_config_yaml()
 
     def load_config_yaml(self):
@@ -113,6 +115,7 @@ class DectectorTrain:
 
         mlf_logger.log_hyperparams(self.config)
         mlf_logger.log_hyperparams({"split_seed": self.seed_n})
+        mlf_logger.log_hyperparams({"cli_args": self.args})
 
         lightning_model = FasterRCNN(self.config)
 
@@ -120,8 +123,8 @@ class DectectorTrain:
             max_epochs=self.config["num_epochs"],
             accelerator=self.accelerator,
             logger=mlf_logger,
-            fast_dev_run=self.args.fast_dev_run,
-            limit_train_batches=self.args.limit_train_batches,
+            fast_dev_run=self.fast_dev_run,
+            limit_train_batches=self.limit_train_batches,
         )
 
         # Run training
--- a/crabs/detection_tracking/train_model.py
+++ b/crabs/detection_tracking/train_model.py
@@ -36,6 +36,7 @@ class DectectorTrain:
     """
 
     def __init__(self, args):
+        self.args = args
         self.config_file = args.config_file
         self.images_dirs = self.prep_img_directories(
             args.dataset_dirs  # args only?
@@ -119,6 +120,8 @@ class DectectorTrain:
             max_epochs=self.config["num_epochs"],
             accelerator=self.accelerator,
             logger=mlf_logger,
+            fast_dev_run=self.args.fast_dev_run,
+            limit_train_batches=self.args.limit_train_batches,
         )
 
         # Run training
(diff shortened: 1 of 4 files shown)


crabs/detection_tracking/train_model.py
```diff
@@ -36,6 +36,7 @@ class DectectorTrain:
     """
 
     def __init__(self, args):
+        self.args = args
```
sfmig, line 39: fantastic additions! 🚀 

Just about the syntax:
I was wondering if we could have something more descriptive than `args`. Maybe:
```
 def __init__(self, args):
        [...]
	self.debug_fast_dev_run = args.fast_dev_run
	self.debug_limit_train_batches = args.limit_train_batches
	self.load_config_yaml()
```

Pros: it makes it more explicit what we are passing (hence more readable), it would be consistent with the rest of the `args` we have, and we avoid repetition (as it is now, we have `self.args.accelerator` and `self.accelerator` right?). Cons: less flexible when we add new args. But I think pros outweight cons.

crabs/detection_tracking/train_model.py
```diff
@@ -119,14 +120,17 @@ def train_model(self):
             max_epochs=self.config["num_epochs"],
             accelerator=self.accelerator,
             logger=mlf_logger,
+            fast_dev_run=self.args.fast_dev_run,
+            limit_train_batches=self.args.limit_train_batches,
```
sfmig, line 124:
```suggestion
            fast_dev_run=self.debug_fast_dev_run,
            limit_train_batches=self.debug_limit_train_batches,
```

sfmig: Question: do we want these parameters in the config instead? I personally like them as CLI arguments but I'm not sure we have a clear distinction for what goes in config and what doesn't... 

The big advantage of putting it in the config is that all the config is dumped into an MLflow log. Could we also dump the CLI arguments to an MLflow log? I think this would be very useful. We can discuss this offline or open an issue.

crabs/detection_tracking/train_model.py
```diff
@@ -119,14 +120,17 @@ def train_model(self):
...
         )
 
         # Run training
         trainer.fit(lightning_model, data_module)
 
         # Save model if required
         if self.config["save"]:
-            save_model(lightning_model)
+            model_filename = save_model(lightning_model)
+            mlf_logger.log_hyperparams({"model_filename": model_filename})
```
sfmig, line 133: fantastic saving the model name ✨

crabs/detection_tracking/train_model.py
```diff
@@ -191,6 +195,20 @@ def train_parse_args(args):
         default=42,
         help="seed for dataset splits",
     )
+    parser.add_argument(
+        "--fast_dev_run",
+        action="store_true",
+        help="option to run only one batch and one epoch for debugging",
```
sfmig, line 201:
```suggestion
        help="Debugging option to run training for one batch and one epoch",
```

crabs/detection_tracking/train_model.py
```diff
@@ -191,6 +195,20 @@ def train_parse_args(args):
...
+    )
+    parser.add_argument(
+        "--limit_train_batches",
+        type=float,
+        default=1.0,
+        help=(
+            "option to run smaller training number per epoch for debugging."
```
sfmig, line 208:
```suggestion
            "Debugging option to run training on a fraction of the training set."
```

sfmig (approved):
thanks, looking neat! 🚀 some suggestions and one question

crabs/detection_tracking/train_model.py
nikk-nikaznan: I think technically, things like hyper parameters in config, things we need to change quickly and many times in cli. Like debugging so we don't keep changing the config files for example from True to False too much. I don't know really, I think it's preference as well.

nikk-nikaznan: I think we keep as it is for now, we can always move things around especially if we see thing growing too long

PR #146, opened 2024-04-05
Save model checkpoint regularly during training

sfmig:


nikk-nikaznan (approved):
Looking good. I modified the `evaluate_model.py` to reflect the change. It works fine now to load the checkpoint as well.

PR #147, opened 2024-04-05
Compute test and validation splits in one go using the same transforms

sfmig:
Right now we first need to create the dataset and then split it. The transforms are unfortunately defined when creating the dataset (there may be ways around this but haven't digged into it).

Since validation and test set will always have the same transforms, we reduce the repeated dataset generation by computing them in one go.


nikk-nikaznan:
Yes, I am inclining towards this approach. Always use the same transform for both.

nikk-nikaznan (approved):
👍

PR #154, opened 2024-04-10
Remove DEFAULT_ANNOTATIONS_FILENAME definition in train script

sfmig:
It should be imported from detection_utils (but it's actually not used here so I just deleted it)


nikk-nikaznan (approved):
Thank you for spotting that

PR #157, opened 2024-04-11
fix slurm ID for array jobs

sfmig:
Two fixes:
- Fix slurm ID for array jobs: I was using the individual ID rather than the parent ID
- Fix artifact from rebase: I rebased to main using Github interface and didn't check carefully the result.... the logging code was duplicated, I thought I was going crazy 😅


nikk-nikaznan (approved):
Good catch! This looking good!

PR #166, opened 2024-04-16
Refactor mlflow logger setup to re-use in evaluation

sfmig:
This PR moves some of the MLflow logger setup functions to `detection_utils`, so that they are shared in the training and eval script.

### DIFF
--- a/crabs/detection_tracking/evaluate_model.py
+++ b/crabs/detection_tracking/evaluate_model.py
@@ -53,6 +53,7 @@ class DetectorEvaluation:
         self.seed_n = args.seed_n
         self.ious_threshold = args.ious_threshold
         self.score_threshold = args.score_threshold
+        self.mlflow_folder = args.mlflow_folder
         self.load_config_yaml()
 
     def load_config_yaml(self):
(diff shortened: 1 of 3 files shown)


crabs/detection_tracking/evaluate_model.py
```diff
@@ -51,52 +53,76 @@ def __init__(
...
+        Setup MLflow logger for testing.
+
+        Includes logging metadata about the job (CLI arguments and SLURM job IDs).
+        """
+        # Assign run name
+        self.set_run_name()
+
+        # Setup logger (no checkpointing)
+        mlf_logger = setup_mlflow_logger(
+            experiment_name="Sep2023_evaluation",
```
nikk-nikaznan, line 77: Should we have experiment_name this as args?

nikk-nikaznan (approved):

crabs/detection_tracking/evaluate_model.py
sfmig: yes! I am changing the CLI in a separate PR to better match the training one

PR #168, opened 2024-04-18
Fix MLflow folder error

sfmig:
This PR fixes the utf-8 encoding error we were getting with MLflow logger. 

The path passed needs to be a 'raw' string (the forward slashes need to be escaped).

I also changed the default location for the `ml-runs` to be under the current working directory. When we launch a job in the cluster (using `run_training....sh`), I set the MLflow folder argument to be the one we agreed in ceph. Let me know thoughts, I think this would be the most convenient.

I checked this fix works: locally, with an interactive node, with a batch job.


nikk-nikaznan (approved):
works fine for me as well

PR #173, opened 2024-04-18
Log path to checkpoints + cosmetic bits

sfmig:
- log path to checkpoints directory -- a bit rudimentary but ok for now?
- add `limit_val_batches`
- add mps fix for all dataloaders 
- cosmetic edits to class attributes in init


nikk-nikaznan (approved):
Looks good!

PR #175, opened 2024-04-18
Use default config in bash scripts for training

sfmig:
I also modified the reference config to show sensible parameters for training in the cluster.

I wasn't sure about workers, let me know thoughts!


nikk-nikaznan (approved):
I think everything seems good, thank you!. I would set the default num_workers for clusters to be 4. I think num_workers of 0 usually used only for debugging purposes

PR #176, opened 2024-04-19
Training in the cluster guide

sfmig:
A quick guide for us and the final user (probably Sanna) on how to launch the training of a model in the cluster.

### DIFF
--- a/guides/TrainingModelsHPC.md
+++ b/guides/TrainingModelsHPC.md
@@ -40,8 +40,8 @@
 
     When launching a single or an array job, one may want to edit in the bash script:
 
-    - The `--experiment_name` (line 111). This should roughly reflect the reason for running the experiment (for example, include `data_augm` for a data augmentation ablation study). Otherwise we use the name of the dataset (Sep2023).
-    - The `mlflow_folder` (line 113). By default, we point to the "scratch" folder. This holds runs that we don't need to keep. For runs we would like to consider for evaluation, we will instead point to the folder at `/ceph/zoo/users/sminano/ml-runs-all/ml-runs`.
+    - The `EXPERIMENT_NAME`. This should roughly reflect the reason for running the experiment (for example, include `data_augm` for a data augmentation ablation study). Otherwise we use the name of the dataset (Sep2023).
+    - The `MLFLOW_FOLDER`. By default, we point to the "scratch" folder at `/ceph/zoo/users/sminano/ml-runs-all/ml-runs-scratch` . This folder holds runs that we don't need to keep. For runs we would like to consider for evaluation, we will instead point to the folder at `/ceph/zoo/users/sminano/ml-runs-all/ml-runs`.
 
     Less frequently, one may need to edit:
 
--- a/guides/TrainingModelsHPC.md
+++ b/guides/TrainingModelsHPC.md
@@ -10,22 +10,113 @@
 
    It may ask for your password twice. To set up SSH keys for the SWC cluster, see [this guide](https://howto.neuroinformatics.dev/programming/SSH-SWC-cluster.html#ssh-keys)
 
-2. Clone this repository somewhere in `ceph` (e.g., in your scratch partition):
+2. Fetch the bash script for training from the repository
+
+   To do so, run any of the following commands. They will place the bash script for training in the current working directory. The files retrieved will be the versions in the `main` branch of the repository.
+
+   To train a single job: fetch the `run_training_single.sh` file
 
    ```
-   cd /ceph/scratch/sminano
-   git clone https://github.com/SainsburyWellcomeCentre/crabs-exploration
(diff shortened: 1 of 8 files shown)


guides/TrainingModelsHPC.md
```diff
@@ -0,0 +1,169 @@
...
+    - If not in the SWC network: connect to the SWC VPN.
+
+1.  **Connect to the SWC HPC cluster**
+
+    ```
+    ssh hpc-gw1
+    ```
+
+    It may ask for your password twice. To set up SSH keys for the SWC cluster, see [this guide](https://howto.neuroinformatics.dev/programming/SSH-SWC-cluster.html#ssh-keys).
+
```
nikk-nikaznan, line 14: Should we add line to git clone and setup env as the first step, or maybe prep step?

guides/TrainingModelsHPC.md
```diff
@@ -0,0 +1,169 @@
...
+1.  **Fetch the training script from the 🦀 repository**
```
nikk-nikaznan, line 15:
```suggestion
2.  **Fetch the training script from the 🦀 repository**
```

guides/TrainingModelsHPC.md
```diff
@@ -0,0 +1,169 @@
...
+
+    To do so, run any of the following commands. They will place the bash script for training in the current working directory. The files retrieved will be the versions in the `main` branch of the repository.
```
nikk-nikaznan, line 17: Do we need to fetch? If they have the latest version of the repo, then it is technically there. Or is it to ensure the changes in the bash script being replace by fetching the file?

guides/TrainingModelsHPC.md
```diff
@@ -0,0 +1,169 @@
...
+      ```
+
+    To retrieve a different version of any of these files, from the one at the tip of `main`:
+
+    - replace `main` in the remote file path after `curl` with the relevant branch, or
+    - replace `main` with `blob/commit-hash` to retrieve a specific commit.
+
+    These bash scripts fetch the crabs package from git, install it in the compute node, and run a training job.
+
+1.  **Edit the training bash script if required**
```
nikk-nikaznan, line 37:
```suggestion
3.  **Edit the training bash script if required**
```

guides/TrainingModelsHPC.md
```diff
@@ -0,0 +1,169 @@
...
+
+    - the `DATASET_DIR`: usually we train on the "Sep2023_labelled" dataset;
+    - the `TRAIN_CONFIG_FILE`: usually we point to the file at `/ceph/zoo/users/sminano/cluster_train_config.yaml` which we can edit. Note that all config parameters are logged in MLflow;
+    - the `GIT_BRANCH`, if we want to use a specific version of the package. Usually we will run the version in `main`.
+
+    Additionally for an array job, one may want to edit the number of jobs in the array (by default set to 3):
+
+    - if the number of jobs in the array is edited, the `LIST_SEEDS` needs to be modified accordingly, otherwise we will get an error when launching the job.
+
+1.  **Edit the config if required**
```
nikk-nikaznan, line 56:
```suggestion
4.  **Edit the config if required**
```

guides/TrainingModelsHPC.md
```diff
@@ -0,0 +1,169 @@
...
+
+5.  **Run the training job using the SLURM scheduler**
+
+    To launch a job, use the `sbatch` command with the training script you fetched (and maybe edited):
+
+    ```
+    sbatch <path-to-fetched-training-bash-script>
+    ```
+
+1.  **Check the status of the training job**
```
nikk-nikaznan, line 88:
```suggestion
6.  **Check the status of the training job**
```

nikk-nikaznan (approved):
Seems good to me. Just some questions mostly.

guides/TrainingModelsHPC.md
sfmig: We don't need to do that because the bash script installs the crabs package from git

guides/TrainingModelsHPC.md
sfmig: Note that is not `git fetch`, but just downloading the file with `curl` - will change it to "download" for clarity.

It is easier than git cloning the repo to the login node, since we only need the bash scripts. The actual 🦀 package is installed in the compute node when running the bash script.

guides/TrainingModelsHPC.md
sfmig: yeah sorry, in markdown if you label all items in an enumeration with `1., 1., 1.` it will reformat them as `1., 2., 3.` (with the advantage that you don't have to keep track of them all the time). But when I use alerts (i.e., Caution or Tip blocks), they need to be left-aligned and so that resets the numbering of the enumeration.

I changed it so that numbers are restarted after the tip/caution blocks!

PR #180, opened 2024-06-04
Saving mota output

nikk-nikaznan:
- mota return all the metrics and save to csv
- adding some plot functions to visualise the tracking result

-----
NOTES by @sfmig:
I split this PR in smaller parts:
- a refactoring PR #241 
- a small refactoring to the detector evaluation  in #240 
- adding MLflow logger and using Lightning's predict methon in #238 
- a MOTA histogram PR #239 
- a tracker evaluation review & fix in PR #243 (once the first two are merged)


crabs/tracker/utils/io.py
```diff
@@ -154,6 +156,7 @@ def save_required_output(
         frame_copy = frame.copy()
         for bbox in tracked_boxes:
             xmin, ymin, xmax, ymax, id = bbox
+            print(f"Calling draw_bbox with {bbox}")
```
sfmig, line 159:
```suggestion
         
```

crabs/tracker/utils/io.py
```diff
@@ -178,3 +181,149 @@ def release_video(video_output) -> None:
...
+        for row in reader:
+            true_positives_list.append(int(row["True Positives"]))
+            missed_detections_list.append(int(row["Missed Detections"]))
+            false_positives_list.append(int(row["False Positives"]))
+            num_switches_list.append(int(row["Number of Switches"]))
+            total_ground_truth_list.append(int(row["Total Ground Truth"]))
+            mota_value_list.append(float(row["Mota"]))
+
+    return (
+        true_positives_list,
```
sfmig, line 220: maybe this tuple can be a dict instead? It's a bit less of a code smell

sfmig: I think if we read the csv as a pandas dataframe instead we can extract the columns more efficiently (that is, without explicit looping).

There is also a dataframe [`.to_dict()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_dict.html) method, so we may be able to get the output dictionary in one go this way.

crabs/tracker/utils/tracking.py
```diff
@@ -152,3 +153,12 @@ def prep_sort(prediction: dict, score_threshold: float) -> np.ndarray:
...
 
     return np.asarray(pred_sort)
+
+
+def save_tracking_mota_metrics(
+    tracking_output_dir: Path,
+    track_results: dict[str, Any],
+) -> None:
+    track_df = pd.DataFrame(track_results)
+    output_filename = f"{tracking_output_dir}/tracking_metrics_output.csv"
```
sfmig, line 163: Just to follow our usual convention, could we timestamp the output directory? We would need to ensure it goes in the same directory as the video output if requested (or other outputs that may be requested)

sfmig (changes requested):
thanks Nik! 🚀 

Some small suggestions, and a slightly larger one re the output directory naming.

crabs/tracker/utils/tracking.py
sfmig: Fixed in #229

PR #181, opened 2024-06-05
MOTA revisited 

nikk-nikaznan:
- The old way of getting the ID switch is by finding the symmetric_difference between two list. The problem with this if we have `prev_id = [1, 2, 3]` and `current_id = [1, 2, 4]` the `n_switches =2`. 
- The new approach will be 1. 
- We also now have more check on the id list. So if  `prev_id = [1, 2, 3]` and `current_id = [1, 2]` with `total_gt = 2`, `n_switches = 0`. 
-  But,  if  `prev_id = [1, 2, 3]` and `current_id = [1, 2]` with `total_gt = 3`, `n_switches = 0` because this is missed_detection.
- Similarly,  if  `prev_id = [1, 2, 3]` and `current_id = [1, 2, 3, 4]` with `total_gt = 3`, `n_switches = 0` because this is false_positive.
- Because of that, we now only considering the IDs that have been matched to ground truth as `current_id` to avoid double counting. 
- Some changes for the test -- feel free to play around with the value in the test to see any other possibility.

### DIFF
--- a/crabs/detection_tracking/tracking_utils.py
+++ b/crabs/detection_tracking/tracking_utils.py
@@ -126,35 +126,36 @@ def evaluate_mota(
     - Identity Switches: Instances where the tracking algorithm assigns a different ID to an object compared to its ID in the previous frame.
     - Total Ground Truth: The total number of ground truth objects in the scene.
 
-    The MOTA score ranges from 0 to 1, with higher values indicating better tracking performance.
+    The MOTA score ranges from -inf to 1, with higher values indicating better tracking performance.
     A MOTA score of 1 indicates perfect tracking, where there are no missed detections, false positives, or identity switches.
     """
     total_gt = len(gt_boxes)
     false_positive = 0
+    matched_gt_boxes = set()
+    matched_tracked_boxes = set()
 
     for i, tracked_box in enumerate(tracked_boxes):
         best_iou = 0.0
         best_match = None
 
         for j, gt_box in enumerate(gt_boxes):
-            iou = calculate_iou(gt_box[:4], tracked_box[:4])
-            if iou > iou_threshold and iou > best_iou:
-                best_iou = iou
-                best_match = j
+            if j not in matched_gt_boxes:
+                iou = calculate_iou(gt_box[:4], tracked_box[:4])
+                if iou > iou_threshold and iou > best_iou:
+                    best_iou = iou
+                    best_match = j
+
         if best_match is not None:
             # successfully found a matching ground truth box for the tracked box.
-            # set the corresponding ground truth box to None.
-            gt_boxes[best_match] = None
+            matched_gt_boxes.add(best_match)
+            matched_tracked_boxes.add(i)
         else:
             false_positive += 1
 
-    missed_detections = 0
-    for box in gt_boxes:
-        if box is not None and not np.all(np.isnan(box)):
-            # if true ground truth box was not matched with any tracked box
-            missed_detections += 1
+    missed_detections = total_gt - len(matched_gt_boxes)
 
-    tracked_ids = [[box[-1] for box in tracked_boxes]]
+    tracked_ids = [[tracked_boxes[i][-1] for i in matched_tracked_boxes]]
+    print(tracked_ids)
 
     num_switches = count_identity_switches(
         prev_frame_ids, tracked_ids, total_gt
--- a/crabs/detection_tracking/tracking_utils.py
+++ b/crabs/detection_tracking/tracking_utils.py
@@ -52,6 +52,7 @@ def calculate_iou(box1: np.ndarray, box2: np.ndarray) -> float:
 def count_identity_switches(
     prev_frame_ids: Optional[list[list[int]]],
     current_frame_ids: Optional[list[list[int]]],
+    n_gt: int,
 ) -> int:
     """
     Count the number of identity switches between two sets of object IDs.
@@ -72,16 +73,19 @@ def count_identity_switches(
     if prev_frame_ids is None or current_frame_ids is None:
         return 0
 
-    # Initialize count of identity switches
-    num_switches = 0
-
     prev_ids = set(prev_frame_ids[0])
     current_ids = set(current_frame_ids[0])
 
-    # Calculate the number of switches by finding the difference in IDs
-    num_switches = len(prev_ids.symmetric_difference(current_ids))
+    # number of ID in current frame that is not in previous frame
+    n_difference = len(list(current_ids - prev_ids))
 
-    return num_switches
+    if prev_ids != current_ids:
+        if (len(current_ids) == n_gt) and (n_difference == 0):
+            return 0
+        else:
+            return n_difference
+    else:
+        return 0
 
 
 def evaluate_mota(
@@ -152,7 +156,9 @@ def evaluate_mota(
 
     tracked_ids = [[box[-1] for box in tracked_boxes]]
 
-    num_switches = count_identity_switches(prev_frame_ids, tracked_ids)
+    num_switches = count_identity_switches(
+        prev_frame_ids, tracked_ids, total_gt
+    )
 
     mota = 1 - (missed_detections + false_positive + num_switches) / total_gt
     return mota
--- a/tests/test_unit/test_tracking_utils.py
+++ b/tests/test_unit/test_tracking_utils.py
@@ -17,26 +17,23 @@ from crabs.detection_tracking.tracking_utils import (
 
 
 @pytest.mark.parametrize(
-    "prev_frame_id, current_frame_id, expected_output",
+    "prev_frame_id, current_frame_id, n_gt, expected_output",
     [
-        (None, [[6, 5, 4, 3, 2, 1]], 0),
-        (
-            [[6, 5, 4, 3, 2, 1]],
-            [[6, 5, 4, 3, 2, 1]],
-            0,
-        ),  # no identity switches
-        ([[5, 6, 4, 3, 1, 2]], [[6, 5, 4, 3, 2, 1]], 0),
-        ([[6, 5, 4, 3, 2, 1]], [[6, 5, 4, 2, 1]], 1),
-        ([[6, 5, 4, 2, 1]], [[6, 5, 4, 2, 1, 7]], 1),
-        ([[6, 5, 4, 2, 1, 7]], [[6, 5, 4, 2, 7, 8]], 2),
-        ([[6, 5, 4, 2, 7, 8]], [[6, 5, 4, 2, 7, 8, 3]], 1),
+        (None, [[6, 5, 4, 3, 2, 1]], 6, 0),
+        ([[6, 5, 4, 3, 2, 1]], [[6, 5, 4, 3, 2, 1]], 6, 0),
+        ([[5, 6, 4, 3, 1, 2]], [[6, 5, 4, 3, 2, 1]], 6, 0),
+        ([[6, 5, 4, 3, 2, 1]], [[6, 5, 4, 2, 1]], 5, 0),
+        ([[6, 5, 4, 3, 2, 1]], [[6, 5, 4, 2, 1, 7]], 6, 1),
+        ([[6, 5, 4, 2, 1]], [[6, 5, 4, 2, 7]], 5, 1),
+        ([[6, 5, 4, 3, 2]], [[6, 5, 4, 2, 7, 8, 3]], 5, 2),
+        ([[6, 5, 4, 3]], [[6, 5, 4, 3, 7, 8, 9]], 5, 3),
     ],
 )
 def test_count_identity_switches(
-    prev_frame_id, current_frame_id, expected_output
+    prev_frame_id, current_frame_id, n_gt, expected_output
 ):
     assert (
-        count_identity_switches(prev_frame_id, current_frame_id)
+        count_identity_switches(prev_frame_id, current_frame_id, n_gt)
         == expected_output
     )
 
--- a/tests/test_unit/test_tracking_utils.py
+++ b/tests/test_unit/test_tracking_utils.py
@@ -95,15 +95,17 @@ def test_perfect_tracking(gt_boxes, tracked_boxes, prev_frame_ids):
 
 
 def test_missed_detections(gt_boxes, tracked_boxes, prev_frame_ids):
-    # Remove one ground truth box to simulate a missed detection
-    gt_boxes = np.delete(gt_boxes, 0, axis=0)
+    # Remove one tracked box to simulate a missed detection
+    tracked_boxes = np.delete(tracked_boxes, 0, axis=0)
     mota = evaluate_mota(
         gt_boxes,
         tracked_boxes,
         iou_threshold=0.1,
         prev_frame_ids=prev_frame_ids,
     )
+    # mota = 1 - (1 + 0 + 0) / 3
     assert mota < 1.0
+    assert mota == pytest.approx(2 / 3)
 
 
 def test_false_positives(gt_boxes, tracked_boxes, prev_frame_ids):
@@ -115,7 +117,9 @@ def test_false_positives(gt_boxes, tracked_boxes, prev_frame_ids):
         iou_threshold=0.1,
         prev_frame_ids=prev_frame_ids,
     )
+    # mota = 1 - (0 + 1 + 0) / 3
     assert mota < 1.0
+    assert mota == pytest.approx(2 / 3)
 
 
 def test_identity_switches(gt_boxes, tracked_boxes, prev_frame_ids):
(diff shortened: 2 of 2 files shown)


sfmig:
hey @nikk-nikaznan, I was just having a look at this and I get a different result:

> Similarly, if prev_id = [1, 2, 3] and current_id = [1, 2, 3, 4] with current_gt = 4, n_switches = 0. (A new ID is appear in the current frame)

For me this returns `n_switches=1`
image

I think this is not intended right?

sfmig:
I think it would be something like follows:
1. For the bounding boxes that match the ground truth in frame **f** (i.e., that overlap sufficiently with the ground truth box), compute the map (dictionary) from ground truth IDs to predicted IDs.
1. Get the same map for the previous frame **f-1** (we would have stored it).
1. The number of ID switches is the number of times the following thing occurs: for every ground-truth ID (keys of the dictionary), when do the predicted IDs disagree? 

Let me know thoughts, happy to chat offline if you prefer.

crabs/detection_tracking/inference_model.py
```diff
@@ -65,12 +68,11 @@ def load_trained_model(self) -> torch.nn.Module:
         -------
         torch.nn.Module
         """
-        model = torch.load(
-            self.args.model_dir,
-            map_location=torch.device(self.args.accelerator),
-        )
-        model.eval()
-        return model
+        # Get trained model
```
sfmig, line 71: Can we avoid this code repetition? Maybe doing something similar to what we did with optuna.

crabs/detection_tracking/inference_model.py
```diff
@@ -19,6 +20,8 @@
     draw_bbox,
 )
 
+DEVICE = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
sfmig, line 23: should this be set via CLI arguments? (We do that in the other modules right?)

tests/test_unit/test_tracking_utils.py
```diff
@@ -17,26 +17,23 @@
 
 
 @pytest.mark.parametrize(
-    "prev_frame_id, current_frame_id, expected_output",
+    "prev_frame_id, current_frame_id, n_gt, expected_output",
```
sfmig, line 20: The test of an "appearing crab" fails if the number of ID switches is expected to be 0


This is the test I tried:
```
@pytest.mark.parametrize(
    "prev_frame_id, current_frame_id, n_gt, expected_output",
    [
        ([[4, 3, 2, 1]], [[5, 4, 3, 2, 1]], 5, 0),
    ],
)
def test_count_identity_switches(
    prev_frame_id, current_frame_id, n_gt, expected_output
):
    assert (
        count_identity_switches(prev_frame_id, current_frame_id, n_gt)
        == expected_output
    )
```

sfmig (changes requested):
I think the `count_identity_switches` is not behaving as expected 🤔 

I explained in the comments of the PR conversation.

Maybe a good one to pair program? Let me know thoughts

samcunliffe:
> @sfmig  I think there is some confusion, because as I understand it from the [HOTA paper](https://link.springer.com/article/10.1007/s11263-020-01375-2) we only check ID switches for the true positives. So we shouldn't be filtering out missed detections or false negatives inside this function, right?

https://github.com/JonathonLuiten/TrackEval/blob/12c8791b303e0a0b50f753af204249e622d0281a/trackeval/metrics/clear.py#L93-L97

crabs/detection_tracking/tracking_utils.py
```diff
@@ -50,46 +50,50 @@ def calculate_iou(box1: np.ndarray, box2: np.ndarray) -> float:
...
-    num_switches = 0
-
-    prev_ids = set(prev_frame_ids[0])
-    current_ids = set(current_frame_ids[0])
+    switch_count = 0
 
-    # Calculate the number of switches by finding the difference in IDs
-    num_switches = len(prev_ids.symmetric_difference(current_ids))
+    for current_gt_id, current_tracked_id in current_frame_id_map.items():
+        prev_tracked_id = prev_frame_id_map.get(current_gt_id)
```
samcunliffe, line 79:
```suggestion
        prev_tracked_id = prev_frame_id_map.get(current_gt_id)
        prev_gt_id = prev_frame_gt_id_map.get(current_tracked_id)
        
        if current_gt_id != prev_gt_id:
            switch_count += 1
```

samcunliffe: Exercise for the reader: has Sam written total garbage?

tests/test_unit/test_tracking_utils.py
```diff
@@ -19,17 +19,14 @@
...
-        ([[6, 5, 4, 2, 1, 7]], [[6, 5, 4, 2, 7, 8]], 2),
-        ([[6, 5, 4, 2, 7, 8]], [[6, 5, 4, 2, 7, 8, 3]], 1),
+        (None, {1: 1, 2: 2, 3: 3, 4: 4}, 0),
+        ({1: 1, 2: 2, 3: 3, 4: 4}, {1: 1, 2: 2, 3: 3, 4: 4}, 0),
+        ({1: 1, 2: 2, 3: 3, 4: 4}, {1: 2, 2: 1, 3: 3, 4: 4}, 2),
+        ({1: 1, 2: 2, 3: 3, 4: 4}, {1: 1, 2: 2, 3: 3}, 0),
+        ({1: 1, 2: 2, 3: 3}, {1: 1, 2: 2, 3: 3, 4: 4}, 0),
+        ({1: 1, 2: 2, 3: 3}, {1: 1, 2: 2, 4: 3}, 1),
+        ({1: 1, 2: 2, 3: 3, 4: 4}, {1: 1, 2: 2, 3: 3, 4: 5}, 1),
+        ({1: 1, 2: 2, 3: 3, 4: 4}, {1: 1, 2: 2, 3: 3, 4: 5, 5: 6}, 2),
```
samcunliffe, line 29: TODO: if Sam wrote ~OK code above, please add a test for that.

samcunliffe: ... or dive into the HOTA code.

crabs/tracker/evaluate_tracker.py
```diff
@@ -41,35 +44,40 @@ def create_gt_list(
...
                 )
-        return gt_boxes_list
+                gt_ids_list[frame_number] = np.hstack(
+                    [gt_ids_list[frame_number], track_id]
+                )
+
+        return gt_boxes_list, gt_ids_list
 
-    def get_ground_truth_data(self) -> list[np.ndarray]:
+    def get_ground_truth_data(
```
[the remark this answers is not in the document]
sfmig: I think the conclusion is: let's use the frame numbers as they are in the csv (whatever base they are in) rather than frame indices (0-based), and use a dict rather than lists. 

That should make us agnostic to the convention used in the csv (1-based or 0-based, starting in the first frame or not...)

sfmig (changes requested):
thanks Nik! 🚀 

Some bits of the tracking code I hadn't reviewed yet (my bad, I think because of the refactoring?), so I split the review in 2 parts. In this first part, I covered:
- in `evaluate_tracker`, `create_gt_list` and `get_ground_truth_data`
- `track_video`

In the second part I will dive in the id switches definition and the tests.

### output saving
Apart from the points raised in the comments, I have some suggestions on the tracking output. 

Right now both frames and a video are generated when you track an input video. Could we make this optional via CLI arguments? I would say by default we only save the csv with the tracks.

I think we can simplify the structure of the output too. 
- Right now I think the output video is saved two levels down (i.e., `/crabs_track_output/crabs_tracks_label/04.09.2023-04-Right_RE_test_output_video.mp4`). 
- I suggest instead to have under `tracking_output/`:
	- a directory with the video name, e.g. `04.09.2023-04-Right_RE_test_output_video`
	- inside this directory, a subdirectory called `frames` with the pngs (if requested)
	- a video called `tracked_video.mp4`
	- a csv file called `tracks.csv`

I am not fully convinced tho, what do you think?

It would be nice that if we pass the ground truth and request an output video, both detections and ground truth are plotted in the video. This may be too big for this PR tho, so no worries if it's a faff.

crabs/tracker/evaluate_tracker.py
```diff
@@ -131,46 +144,60 @@ def calculate_iou(self, box1: np.ndarray, box2: np.ndarray) -> float:
 
     def count_identity_switches(
         self,
-        prev_frame_ids: Optional[list[list[int]]],
-        current_frame_ids: Optional[list[list[int]]],
+        prev_frame_id_map: Optional[Dict[int, int]],
```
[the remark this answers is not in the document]
nikk-nikaznan: Sorry I should be a bot clear here. This is mapping from ground truth to prediction. I think it is in the docstring

crabs/tracker/track_video.py
```diff
@@ -88,7 +88,7 @@ def load_trained_model(self) -> torch.nn.Module:
             self.args.checkpoint_path
         )
         trained_model.eval()
-        trained_model.to(DEVICE)  # Should device be a CLI?
+        trained_model.to(DEVICE)
         return trained_model
 
     def load_video(self) -> None:
```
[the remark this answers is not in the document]
nikk-nikaznan: > In run_tracking() below, we could combine the two if self.args.gt_path cases together.

I forgot to address this. You had some changes like this before in refactoring. But the evaluation will happen only towards the end, when all the frames have been tracked. I added the gt_path checking at the beginning because it can takes a long time to finish the tracking and only then got error that the path is not exist.

crabs/tracker/track_video.py
```diff
@@ -88,7 +88,7 @@ def load_trained_model(self) -> torch.nn.Module:
             self.args.checkpoint_path
         )
         trained_model.eval()
-        trained_model.to(DEVICE)  # Should device be a CLI?
+        trained_model.to(DEVICE)
```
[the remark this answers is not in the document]
nikk-nikaznan: I think we had a discussion on this last week. Similar to the visualisation, as for inference, it seems using PyTorch way is the simplest approach.

crabs/tracker/track_video.py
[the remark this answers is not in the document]
nikk-nikaznan: > n run_tracking()we set frame_number = 1 ...

Shouldn't be an issue, but good call, I just change to 0 to make it consistent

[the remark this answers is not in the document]
sfmig: > I forgot to address this. You had some changes like this before in refactoring. But the evaluation will happen only towards the end, when all the frames have been tracked. I added the gt_path checking at the beginning because it can takes a long time to finish the tracking and only then got error that the path is not exist.

Ok let's leave it as is, ideally we implement validators at some point - I opened an issue #211

sfmig (approved):
thanks Nik! This is great 🌟 

I reviewed the rest. Apologies that it's a big one 🙏 

The bigger points of this PR are:
- suggested reimplementation of `count_identity_switches` function. 
- `tracked_list` should be a  dictionary (following `gt_data`) rather than a list (otherwise if frames are not valid Python indices it will break).
- dictionaries that map from ground truth ID to predicted ID to include missed detections.
- suggestions to rearrange and remove duplicates in the parametrisation of the ID switches test.

Re the first point, I think you are right and with the latest fixes the suggested algorithm doesn't cover any additional cases compared to the current one (contrary to what we thought initially). 

However, I think it is a bit easier to read and conceptualise, and it allows us to structure the cases to test. Plus if in the future we want to cover that case of reID after a missed detection gap, I think the suggested algorithm would make it easier.

crabs/tracker/utils/io.py
```diff
@@ -29,13 +30,13 @@ def prep_csv_writer(output_dir: str, video_file_root: str):
...
-    crabs_tracks_label_dir = Path(output_dir) / "crabs_tracks_label"
-    tracking_output_dir = crabs_tracks_label_dir / video_file_root
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
+    tracking_output_dir = Path(output_dir + f"_{timestamp}") / video_file_root
     # Create the subdirectory for the specific video file root
     tracking_output_dir.mkdir(parents=True, exist_ok=True)
 
     csv_file = open(
-        f"{str(tracking_output_dir / video_file_root)}.csv",
+        f"{str(tracking_output_dir)}/track_labels.csv",
```
sfmig, line 39: I think a more explicit name could be `tracks` or `bboxes_tracks`.

`labels` makes it sound as it is ground-truth. But here it is the output of the detector+tracking right?

crabs/tracker/track_video.py
```diff
@@ -171,52 +159,58 @@ def run_tracking(self):
...
                 frame,
-                frame_number,
+                frame_idx,
                 pred_scores,
             )
 
             # update frame number
-            frame_number += 1
-
+            frame_idx += 1
```
sfmig, line 209:
```suggestion
            frame_idx += 1
            
```
(just adding a space)

crabs/tracker/track_video.py
```diff
@@ -171,52 +159,58 @@ def run_tracking(self):
...
+        frame_idx = 0
+        self.tracked_bbox_id = []
 
         # Loop through frames of the video in batches
         while self.video.isOpened():
             # Break if beyond end frame (mostly for debugging)
             if (
                 self.args.max_frames_to_read
-                and frame_number > self.args.max_frames_to_read
+                and frame_idx > self.args.max_frames_to_read
```
sfmig, line 171:
```suggestion
                and frame_idx + 1 > self.args.max_frames_to_read
```
or 
```suggestion
                and frame_idx >= self.args.max_frames_to_read
```
Since `self.args.max_frames_to_read` is not 0-based right?

E.g. If we want to read 3 frames max, then `self.args.max_frames_to_read = 3`. But when 3 frames have been read, the index at the start of the 4th iteration is `frame_idx=3`, which is not larger than `self.args.max_frames_to_read`. So we would read a fourth frame, right?

crabs/tracker/track_video.py
```diff
@@ -171,52 +159,58 @@ def run_tracking(self):
...
                 self.args.save_frames,
                 self.tracking_output_dir,
                 self.csv_writer,
                 self.args.save_video,
                 self.video_output,
-                tracked_boxes,
+                tracked_boxes_id_per_frame,
                 frame,
-                frame_number,
+                frame_idx,
```
sfmig, line 204:
```suggestion
                frame_number,
```
I think otherwise there is a mismatch between how we are naming exported frames and how we are reading them.
- When reading we assume the filename contains a frame number (1-based)  -- see line 34 of `crabs/tracker/utils/tracking.py`.
- When exporting we make the filename contain a frame index (0-based).

crabs/tracker/utils/tracking.py
```diff
@@ -31,9 +31,9 @@ def extract_bounding_box_info(row: list[str]) -> Dict[str, Any]:
     height = region_shape_attributes["height"]
     track_id = region_attributes["track"]
 
-    frame_number = int(filename.split("_")[-1].split(".")[0]) - 1
+    frame_idx = int(filename.split("_")[-1].split(".")[0]) - 1
```
sfmig, line 34: Here we assume the filename of the frame has a frame number (1-based) in it. And we compute the `frame_idx` as `frame_number-1`.

This has a few issues in my opinion:
- we are saving `frame_idx` under the key `frame_number`, which would be confusing.
- if the frames are for a video clip that doesn't start with frame 1, then the computed `frame_idx` from the frame won't make sense here.

For a low effort fix I would just have here:
```python
    frame_number = int(filename.split("_")[-1].split(".")[0])
    return {
        "frame_number": frame_number,
        "x": x,
        "y": y,
        "width": width,
        "height": height,
        "id": track_id,
    }
```
And then ensure that whenever we use the value in this dictionary, we use it as a key `frame_number` (1-based) and not as an index `frame_idx` (0-based).

crabs/tracker/evaluate_tracker.py
```diff
@@ -1,93 +1,112 @@
...
-            A list containing ground truth bounding boxes organized by frame number.
+        Dict[int, Dict[str, Any]]:
+            A dictionary where the key is the frame number and the value is another dictionary containing:
+            - 'bbox': A numpy array with shape (N, 4) containing coordinates of the bounding boxes
+            [x, y, x + width, y + height] for every object in the frame.
+            - 'id': A numpy array containing the IDs of the tracked objects.
         """
+        predicted_dict: Dict[int, Dict[str, Any]] = {}
+
+        for frame_idx, frame_data in enumerate(self.predicted_boxes_id):
```
sfmig, line 50:
```suggestion
        for frame_number, frame_data in enumerate(self.predicted_boxes_id):
```

Sorry, maybe I didn't explain myself very well last time re `frame_idx` and `frame_number`. I didn't mean to rename all `frame_number`s to `frame_idx`, since they have a different meaning.

I call `frame_idx` if it's 0-based (aka an index to a list / array). I use `frame_number` if it's 1-based (or basically if we cannot guarantee that it can be used as an index).

Here we need `frame_number` cause it is extracted from the filename. And we don't know if the frame numbers in the files are 0-based or 1-based, but most importantly we don't know if they start in frame 1. If they don't and the first file is `_0343.png` we cannot convert it to an index by subtracting 1, right?

crabs/tracker/evaluate_tracker.py
```diff
@@ -1,93 +1,112 @@
...
+            next(csvreader)  # Skip the header row
+            ground_truth_data = [
+                extract_bounding_box_info(row) for row in csvreader
+            ]
+
+        # Format as a dictionary with key = frame number
+        ground_truth_dict: dict = {}
         for data in ground_truth_data:
-            frame_number = data["frame_number"]
+            frame_idx = data["frame_number"]
```
sfmig, line 83:
```suggestion
            frame_number = data["frame_number"]
```
(because it is not an index in the clip)

crabs/tracker/evaluate_tracker.py
```diff
@@ -1,93 +1,112 @@
...
-                data = extract_bounding_box_info(row)
-                ground_truth_data.append(data)
-                max_frame_number = max(max_frame_number, data["frame_number"])
-
-        # Initialize a list to store the ground truth bounding boxes for each frame
-        gt_boxes_list = [np.array([]) for _ in range(max_frame_number + 1)]
-
-        gt_boxes_list = self.create_gt_list(ground_truth_data, gt_boxes_list)
-        return gt_boxes_list
+            # format as numpy arrays
```
sfmig, line 101:
```suggestion
        # format as numpy arrays
```

crabs/tracker/evaluate_tracker.py
```diff
@@ -131,153 +150,200 @@ def calculate_iou(self, box1: np.ndarray, box2: np.ndarray) -> float:
...
+        # Compute sets of ground truth IDs for current and previous frames
+        gt_ids_current_frame = set(gt_to_tracked_id_current_frame.keys())
+        gt_ids_prev_frame = set(gt_to_tracked_id_previous_frame.keys())
+
+        # Compute lists of ground truth IDs that continue, disappear, and appear
+        gt_ids_cont = list(gt_ids_current_frame & gt_ids_prev_frame)
+        gt_ids_disappear = list(gt_ids_prev_frame - gt_ids_current_frame)
+        gt_ids_appear = list(gt_ids_current_frame - gt_ids_prev_frame)
+
+        # Store used predicted IDs to avoid double counting
```
sfmig, line 185:
```suggestion
        # Store used predicted IDs to avoid double counting
        # In `used_pred_ids` we log IDs from either the current or the previous frame that have been involved in an already counted ID switch.
```

crabs/tracker/evaluate_tracker.py
```diff
@@ -131,46 +115,53 @@ def calculate_iou(self, box1: np.ndarray, box2: np.ndarray) -> float:
...
-        num_switches = len(prev_ids.symmetric_difference(current_ids))
+        for current_gt_id, current_tracked_id in current_frame_id_map.items():
+            prev_tracked_id = prev_frame_id_map.get(current_gt_id)
+            prev_gt_id = prev_frame_gt_id_map.get(current_tracked_id)
+            if prev_tracked_id is not None:
+                if prev_tracked_id != current_tracked_id:
+                    switch_count += 1
+            elif prev_gt_id is not None:
+                if current_gt_id != prev_gt_id:
+                    switch_count += 1
```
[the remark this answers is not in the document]
sfmig: thanks @nikk-nikaznan this is an excellent catch.

I think I agree with example 1 and 2 having an expected switch count = 1

> I have added another condition to see if we already used red_ids in the previous case, then we will not count the switch -- considering we should not count the ID switch on the same predicted ID for more than once.

I see, you used `used_pred_ids` to log if IDs in either the current or the previous frame have been involved in an already counted ID switch. 

I think this makes sense to me! I added a comment to that line just to mega clarify.

PR #182, opened 2024-06-07
Adding checkpoint_path for resume training

### LINKED ISSUE #179: Ability to restart training from a checkpoint
## Definition of done

* The user can restart the training from a checkpoint of a previous training run.
* This is documented i.e. in #148

nikk-nikaznan:
closes #179 
- [x] Checkpoint options
- [x] add checkpoint documentation in the guide


crabs/detection_tracking/train_model.py
```diff
@@ -163,10 +169,46 @@ def core_training(self) -> lightning.Trainer:
...
+        else:
+            trainer.fit(
+                lightning_model,
+                data_module,
+            )
+
+        if checkpoint_callback and checkpoint_callback.last_model_path:
+            logging.info(
+                f"Last checkpoint path: {checkpoint_callback.last_model_path}"
+            )
```
sfmig, line 211: I think we can remove this bit and avoid having to return `checkpoint_callback` in `setup_trainer()`.
```suggestion

```

guides/TrainingModelsHPC.md
```diff
@@ -92,7 +92,25 @@
...
+     - If training is disrupted and stops mid-way, you can resume it by adding `--checkpoint_path $CKPT_PATH \` to your bash script.
+     - The training will pick up from the last saved epoch and continue until the specified n_epoch.
+     - Similarly, if training completes but you want to extend it based on metric evaluations, you can increase the n_epoch value (e.g., from `n` to `n + y`). Again, use `--checkpoint_path $CKPT_PATH \` in your bash script, and training will resume from epoch `n` to `n + y`.
+     - Ensure the `save_weights_only` parameter under `checkpoint_saving` in the config file is set to `False` to resume training, as this option requires loading both weights and the training state.
+
+   - Fine-tunning
+     - This option is useful for fine-tuning a pre-trained model on a different dataset.
+     - It loads the weights from a checkpoint, allowing you to leverage pre-trained weights from another dataset.
+     - Add `--checkpoint_path $CKPT_PATH \` to your bash script to use this option.
+     - Set the `save_weights_only` parameter under `checkpoint_saving` in the config file to `True`, as only the weights are needed for fine-tuning.
```
sfmig, line 111: Some suggestions, feel free to apply as you see fit!

- I would rename the title to "Restarting training from a checkpoint", so that it is not confused with saving checkpoints during training.
- I think the main difference between the two options is that one makes use of a "full" checkpoint, and the other one uses a "weights-only" checkpoint. But you can use both to continue training right? If this is correct, I think we could simplify this part to just say this.
- I suggest moving this whole section to the main README, rather than the HPC bit. What do you think?
- Related to the bug, I think it is important to say here that even if we start training from a checkpoint, the parameters in the yaml config file apply.

sfmig (changes requested):
thanks Nik 🙌 

While reviewing this PR I found a super tricky bug - but I think I found a fix too.

### the bug
Steps to reproduce:
- train a model for one epoch (specifying `n_epochs=1` in the yaml file) and save a `weights_only` checkpoint.
    -  the checkpoint is at the `path_to_checkpoints` parameter logged in MLflow (the name is `last.ckpt`).
- then launch a training job that starts from that checkpoint . Before I launch it, I edit the config file to have `n_epochs=3`.
- In MLflow, this second training job has the same hyperparameters as the job that produced the training (so it has `n_epochs=1` etc), but _in reality_ the job runs for as many epochs as in the yaml file. So it logs `n_epochs=1`, but runs for `n_epochs=3`.

This is a problem because the logged hyperparameters and the actual hyperparameters used don't match.

This does not happen if you restart training from a "full" checkpoint. In that case the .yaml parameters are logged and used.

### the fix
I am still not sure why this is happening - I opened an [issue](https://github.com/SainsburyWellcomeCentre/crabs-exploration/issues/194) to further investigate. But I found we can overwrite the hyperparameters that are logged in the weights-only checkpoint with the yaml ones, by passing the `config` to `load_from_checkpoint`.
```python
lightning_model = FasterRCNN.load_from_checkpoint(
	self.checkpoint_path,
	config=self.config,
)
```

### Question
If we make this changes, whenever we restart training from a ckpt ("full" or "weights-only"), the hyperparameters used will be the .yaml config ones, and not the ckpt ones. 

The question is: would we want the opposite behaviour? (which we can't have now because I don't know how to properly fix this bug). That would be using the hyperparams from the ckpt when loading a ckpt. It seems convenient, but sounds like something one would forget about. Let me know what you think.

I guess this also highlights that we need more detailed tests 😢

crabs/detection_tracking/train_model.py
```diff
@@ -163,10 +169,46 @@ def core_training(self) -> lightning.Trainer:
 
         # Get model
         lightning_model = FasterRCNN(self.config)
-
         # Run training
-        trainer = self.setup_trainer()
-        trainer.fit(lightning_model, data_module)
+        trainer, checkpoint_callback = self.setup_trainer()
```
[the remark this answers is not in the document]
nikk-nikaznan: Thank you, very good and makes more sense for the fine-tuning checkpoint. Otherwise, we create FasterRCNN instance twice.

crabs/detection_tracking/train_model.py
```diff
@@ -94,14 +97,17 @@ def setup_trainer(self):
...
+                max_epochs=self.config["n_epochs"],
+                accelerator=self.accelerator,
+                logger=mlf_logger,
+                enable_checkpointing=enable_checkpointing,
+                callbacks=[checkpoint_callback] if checkpoint_callback else [],
+                fast_dev_run=self.fast_dev_run,
+                limit_train_batches=self.limit_train_batches,
+            ),
+            checkpoint_callback,
         )
```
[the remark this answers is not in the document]
nikk-nikaznan: I thought it is easy to see (and copy) the trained model. But, I guess for real use, logically, one needs to check mlflow anyway.

guides/TrainingModelsHPC.md
nikk-nikaznan: > you can use both to continue training right?
No quite, as the weights-only one is more for fine-tuning or transfer learning and is not necessarily continue, we can change the whole hyperparameter.

> I suggest moving this whole section to the main README, rather than the HPC bit
We can have them in both? We can add them later in README when we update README #192

nikk-nikaznan:
Good job for spotting the bug and thank you for the review.
Sorry I only check the metrics most of the time on mlflow. It makes sense, because in the original code, when `lightning_model = FasterRCNN(self.config)` we created the model instance, all the hyper parameters are saving to mlflow there. When we call `lightning_model = FasterRCNN.load_from_checkpoint(self.checkpoint_path)`, we recreated the instance, but because we didn't pass any config, it will take everything from the checkpoint. And because the save_hyperparameters is there, it will overwrite the config from yaml. Hence this happened only to the `weights-only` and not the `full` one.

I don't see anything wrong with the new implementation you proposed, definitely make sense. It is only happened on the `weights-only` where technically the fine-tuning/transfer learning. So we only care about the new parameters we used to do the fine-tuning. And the hyper parameters from the checkpoint can always be traced back.

For the full one, it will always be based on the checkpoint, as it is resume training. That's why we need the full state. And, if you train for 10 epochs, and not change the yaml to make it more than 10, no training will happening. As it reached the max_epoch.

Hope this makes sense, happy to chat offline.

crabs/detection_tracking/train_model.py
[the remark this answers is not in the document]
sfmig: yeah it's a good point, but I felt it was not worth having an extra thing to return

guides/TrainingModelsHPC.md
sfmig: ok in both sounds good - I made a note in the PR!

sfmig:
thanks for clarifying Nik!

Some questions:
- I think in the "full" one, it continues the training with the number of epochs it left off, but it will use the batch size from the yaml config. I think this is correct right? is this what we expect?
- the thing I found odd with the "weights-only" was that in MLflow it _logged_ the checkpoint's hyperparameters, but it _used_ the config yaml's hyperparameters instead. I am not sure why that happened, do you know?

We can chat later today if it's easier yes!

crabs/detection_tracking/train_model.py
```diff
@@ -99,7 +102,7 @@ def setup_trainer(self):
             accelerator=self.accelerator,
             logger=mlf_logger,
             enable_checkpointing=enable_checkpointing,
-            callbacks=checkpoint_callback,
+            callbacks=[checkpoint_callback] if checkpoint_callback else [],
```
sfmig, line 105: Question: I think we don't need to pass a list here right?
```suggestion
            callbacks= checkpoint_callback,
```

crabs/detection_tracking/train_model.py
```diff
@@ -161,12 +164,58 @@ def core_training(self) -> lightning.Trainer:
...
+                self.checkpoint_path,
+                config=self.config,
+                # overwrite checkpoint hyperparameters with config ones
+                # otherwise ckpt hyperparameters are logged to MLflow, but yaml hyperparameters are used
+            )
+        else:
+            lightning_model = FasterRCNN(self.config)
+
+        # Get trainer
+        trainer = self.setup_trainer()  # this gets yaml!
```
sfmig, line 202:
```suggestion
        trainer = self.setup_trainer()  
```
sorry sometimes I leave my random notes-to-self

crabs/detection_tracking/train_model.py
```diff
@@ -161,12 +164,58 @@ def core_training(self) -> lightning.Trainer:
...
-        trainer = self.setup_trainer()
-        trainer.fit(lightning_model, data_module)
+        # Resume from full checkpoint if available
+        # (automatically restores model, epoch, step, LR schedulers, etc...)
+        # https://lightning.ai/docs/pytorch/stable/common/checkpointing_basic.html#save-hyperparameters
+        if checkpoint_type == "full":
+            trainer.fit(
+                lightning_model,
+                data_module,
+                ckpt_path=self.checkpoint_path,  # needs to be "full" checkpoint right?
```
sfmig, line 212:
```suggestion
                ckpt_path=self.checkpoint_path,  # needs to having been saved with `save_weights_only=False`
```

sfmig (approved):
thanks Nik! 🚀 
Just one question and small suggestions

PR #136, opened 2024-03-26
Nikkna/lr dev

nikk-nikaznan:

PR #188, opened 2024-06-17
Suggestion to refactor optuna code

sfmig:
This PR builds on work from #136.

Contributions:
- reduces code duplication,
- makes the hyperparameters to sweep across optional. Before, it only `learning_rate` or `n_epochs` were passed it would throw an error,
- ~makes the hyperparameters we can sweep across more flexible.~ --> removed since it was potentially less clear
    - ~I am happy to keep in the guide that we only consider `n_epochs` and `learning_rate` for now, and we could just test those, but maybe this approach makes it easier to implement other in the future.~

It may be a good idea to integrate the tests from #136 here.


crabs/detection_tracking/optuna.py
```diff
@@ -0,0 +1,22 @@
...
+def compute_optimal_hyperparameters(
+    objective_fn: Callable, config_optuna: Dict, direction: str = "maximize"
+):
+    # Create an Optuna study
+    study = optuna.create_study(direction=direction)
+
+    # Optimize the objective function
+    study.optimize(
+        objective_fn,  # takes trial, returns float
+        n_trials=config_optuna["n_trials"][0],  # why 0?
```
sfmig, line 15:
```suggestion
        n_trials=config_optuna["n_trials"][0],  
```
why 0?

sfmig: I changed `n_trials` in the config

crabs/detection_tracking/models.py
```diff
@@ -156,17 +156,26 @@ def on_validation_epoch_end(self) -> None:
...
-        self.validation_step_outputs = self.compute_precision_recall_epoch(
+        (
+            self.validation_step_outputs,
+            val_precision,
+            val_recall,
+        ) = self.compute_precision_recall_epoch(
             self.validation_step_outputs, "val"
         )
 
+        self.log("val_precision", val_precision)
```
sfmig, line 167: Question: should we use `self.logger.log_metrics` instead?

crabs/detection_tracking/config/faster_rcnn.yaml
```diff
@@ -28,3 +28,11 @@ gaussian_blur_params:
   sigma:
     - 0.1
     - 5.0
+optuna:
+  n_trials: 3
```
sfmig, line 32: I made `n_trials` a scalar rather than a list

crabs/detection_tracking/train_model.py
```diff
@@ -116,7 +118,67 @@ def setup_trainer(self):
...
+            The trial to optimise.
+
+        Returns
+        -------
+        float
+            The value to maximise.
+        """
+        # Sample hyperparameters from the search space for this trial
+        # learning rate (a float)
+        if "learning_rate" in self.config["optuna"]:
```
sfmig, line 141: It would be nice if we could be flexible for any hyperparameter defined in `self.config['optuna']`.

So we could say: if the hyperparameter is an integer, use `trial.suggest_int(..)`. But if it is a float, use `trial.suggest_float(..)`.

Can we do this?

crabs/detection_tracking/train_model.py
```diff
@@ -116,7 +118,67 @@ def setup_trainer(self):
...
+
+        # Run training
+        trainer = self.core_training()
+
+        # Return metric to maximise
+        val_precision = trainer.callback_metrics["val_precision"].item()
+        val_recall = trainer.callback_metrics["val_recall"].item()
+        return (val_precision + val_recall) / 2
+
+    def core_training(self) -> lightning.Trainer:
```
sfmig, line 174: suggestions for a better name for this function are welcome 😬

crabs/detection_tracking/train_model.py
```diff
@@ -116,7 +118,67 @@ def setup_trainer(self):
...
+                )
+            else:
+                self.config["learning_rate"] = trial.suggest_float(
+                    "learning_rate",
+                    float(self.config["optuna"]["learning_rate"][0]),
+                    float(self.config["optuna"]["learning_rate"][1]),
+                )
+
+        # num_epochs (an integer)
+        if "num_epochs" in self.config["optuna"]:
```
sfmig, line 154: this check that both bounds are defined feels like it belongs somewhere else....

we can also remove it 🤔

crabs/detection_tracking/train_model.py
```diff
@@ -116,7 +121,68 @@ def setup_trainer(self):
...
+        ----------
+        trial : optuna.Trial
+            The trial to optimise.
+
+        Returns
+        -------
+        float
+            The value to maximise.
+        """
+        # Sample hyperparameters from the search space for this trial
```
sfmig, line 142: I tried to make it a bit more general but let me know thoughts, not fully convinced 🤔

crabs/detection_tracking/models.py
nikk-nikaznan: logger.log_metrics not working for trainer.callback_metrics which what this log for, to be used during optuna trial.

crabs/detection_tracking/train_model.py
nikk-nikaznan: I found this is confusing and longer 😅

crabs/detection_tracking/train_model.py
nikk-nikaznan: The reason the optuna function is not in train_model is to avoid this file getting too long. I feel like main file, should be short

sfmig: yeah agreed, I only kept the bits that are very specific to this training and kept the general optuna parts in the optuna module, but happy to discuss.

crabs/detection_tracking/models.py
sfmig: aaah ok. Maybe we make a if-optuna case to make it clear?

crabs/detection_tracking/optuna.py
```diff
@@ -0,0 +1,56 @@
+from typing import Callable, Dict, Union
+
+import optuna
```
nikk-nikaznan, line 3: I don't think we can call this file, optuna. As the library is called optuna. We can rename this file as `hpo.py` which I think that what people used nowadays

crabs/detection_tracking/train_model.py
```diff
@@ -116,7 +122,68 @@ def setup_trainer(self):
             limit_train_batches=self.limit_train_batches,
         )
 
-    def train_model(self):
+    def optuna_maximise_val_precision_recall(
```
nikk-nikaznan, line 125: Can we call it objective, please? 
It's very confusing when it is not called objective. (that's why I prefer it to be in it's own file).
We have the explanation of what the function do in the docstring.

crabs/detection_tracking/train_model.py
nikk-nikaznan: What is wrong with the one before? I think having optuna objective in `train_model.py` will just make it messy, as optuna option is an enhancement. It is not something people use frequently (as it takes at least a week to get decent study done).

crabs/detection_tracking/train_model.py
nikk-nikaznan: It was much clearer when we know exactly what hyperparameter we have for optuna

nikk-nikaznan (changes requested):
This caused error. We cannot rename the file `optuna.py` as that what the library is called. We can called it `hpo.py` I think that is very common filename in ML space

crabs/detection_tracking/train_model.py
sfmig: I agree optuna is an enhancement and people won't use it frequently. In this PR I was trying to reduce [this code duplication](https://github.com/SainsburyWellcomeCentre/crabs-exploration/blob/d3f17e41f0326d81910a287c521936d7bc27116e/crabs/detection_tracking/optuna_fn.py#L83-L106).

My thinking was that it is more confusing to have repeated parts of the training code in different modules.

crabs/detection_tracking/train_model.py
sfmig: yes, it may be more clear but it is also somewhat "hardcoded" right? With this approach we can extend it to any float or int hyperparameter. 

I personally prefer this one but it is more important that we both find this code clear so will revert to having the hyperparameters defined explicitly. Let me know if you disagree.

crabs/detection_tracking/optuna.py
sfmig: I asked the Python hivemind about this, it shouldn't give an error since the namespace is different but we may want to prevent any issues by renaming the module to a private `_optuna`. 

I'll do that for now, but when we chat we can also try debugging why that is happening for you.

nikk-nikaznan (approved):
Happy to merge this. Should we do the test in the next PR as the test is already there.

Issue #186, opened 2024-06-14
Run inference on the cluster


sfmig:
Streamline a bit the process of running inference on new videos using a trained model on the hpc.

Similar to #163 and the work done to run training on the cluster.

Including a quick guide if possible.

samcunliffe:
Creating environment running into permissions errors...?

samcunliffe:
[@sfmig to review the PR]

Inference:
- Producing tracks from a video.
- Can pass GT to inference to visualise GT vs tracking.
   - if you do this you might as well run the evaluation metrics(?)

What should we log? MLFlow?
- We don't always have a MOTA value.
- If we don't choose to save frames & don't have GT ... we should default to saving the CSV (but it's not necessarily log-able).

samcunliffe:
Idea:

`crabs.tracking`
  - `crabs.tracking.inference`
  - `crabs.tracking.evaluation`

`crabs.detection`
 -  ❓ `crabs.detection.inference`
  - `crabs.detection.evaluation`

sfmig:
Implemented in #251 with the new updates to the CLI and used to generate tracked videos.

It would need to be updated once MLflow is implemented for the detect-and-track command (see #254) - but we can close this issue now as that is now captured in #254

PR #189, opened 2024-06-21
Run Inference on cluster

nikk-nikaznan:
- Bash scripts to run tracking on single video #186 
- Bash script to loop over all escape events for #213 
- Guide on how to run tracker on cluster
- Have tested this on cluster pointed to this branch, these bash scripts have been edited to pointed back to main


crabs/tracker/utils/io.py
```diff
@@ -29,11 +31,15 @@ def prep_csv_writer(output_dir: str, video_file_root: str):
     Tuple
         A tuple containing the CSV writer, the CSV file object, and the tracking output directory path.
     """
-
-    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
-    tracking_output_dir = Path(output_dir + f"_{timestamp}") / video_file_root
-    # Create the subdirectory for the specific video file root
-    tracking_output_dir.mkdir(parents=True, exist_ok=True)
+    if run_on_video_dir:
```
sfmig, line 34: why do we need two separate cases here? Is it because if we run it on a directory there is no `video_file_root`?

Even if we run inference on a directory with videos, it would be nice that the output is timestamped so that we don't overwrite things right?

Maybe we can replace `video_file_root` with the name of the input directory if a directory of videos is passed.

sfmig (changes requested):
thanks @nikk-nikaznan and sorry for the delay!

Just two small suggestions.

The first one is to infer whether the path to the videos is a file or a directory from the path itself, rather than passing that with an extra CLI argument. We can do this easily with `pathlib`'s `is_dir()` or `is_file()`.

The second suggestion is to keep the same output filenaming approach if we pass a directory of videos or a single video as input. Right now, if we pass a directory of videos, the output tracked videos are in a directory which is not timestamped. This is different to the behaviour when we pass a single video and will also overwrite the data if we run the command twice. I gave a suggestion in the comments.

If you have time to implement these changes that would be great! If not let me know and I will do it no problem.

crabs/tracker/utils/io.py
nikk-nikaznan: So, I added this because otherwise we will created 200+ folders in the parent directory. Each video will have tracking_output_timestamp as the creating. of the timestamp is here. Hence I made another loop in the bas script itself.
What I tried to do here is to make one timestamp for this run, within the tracking_output_timestamp we will have folder for each videos. I think that makes more sense and easy to look as well.

sfmig: ah ok that sounds good! thanks!

crabs/tracker/track_video.py
```diff
@@ -299,6 +308,11 @@ def tracking_parse_args(args):
         action="store_true",
         help="Save frame to be used in correcting track labelling",
     )
+    parser.add_argument(
```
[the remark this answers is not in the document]
nikk-nikaznan: @sfmig Just comeback to this. The directory is in the bash, we only pass each video path as args.  I am not quite sure what's the best way to do unless we want to do the loop of the bash in the code. But doesn't seem to be neater

sfmig:
Closing because superseded by #251, which incorporates the changes of #253 (and previously #241)

PR #192, opened 2024-06-24
Update README

sfmig:
Add installation instructions and main commands.

Separate local and hpc training, maybe make two guides?


sfmig:
closing as this work became part of #229

PR #106, opened 2024-02-26
Modules reorganisation

sfmig:
A draft structure for the current detector / tracking modules

PR #193, opened 2024-06-24
Refactor tracking

### LINKED ISSUE #190: Match modules better to tasks
> Idea:
> 
> `crabs.tracking`
>   - `crabs.tracking.inference`
>   - `crabs.tracking.evaluation`
> 
> `crabs.detection`
>  -  ❓ `crabs.detection.inference`
>   - `crabs.detection.evaluation` 

 _Originally posted by @samcunliffe in [#186](https://github.com/SainsburyWellcomeCentre/crabs-exploration/issues/186#issuecomment-2182893860)_

nikk-nikaznan:
After #141 
Another attempt of #106
Probably first attempt towards #190


sfmig:
Hi @nikk-nikaznan,

I liked this organisation, feels clearer! 

I had a go at changing a few things - below an attempt at explaining my thinking. Let me know thoughts.

### general
- I renamed the `detection` module ---> `detector`.
- I renamed the `tracking` module ---> `tracker`.

This is to avoid repetition between higher and lower level modules.

### detector module
- I added a `crabs.detector.utils` submodule.
    - It holds the previous `_utils` modules.
    - Specifically, I split `detection_utils` and `evaluation_utils` into:
        - `crabs.detector.utils.detection`: utils shared by evaluation and training, 
        - `crabs.detector.utils.train`: utils only used in training,
        - `crabs.detector.utils.evaluate`: utils only used in evaluation.
   - I also moved the `optuna` utils there (named `optuna` for now, but we can discuss!) and the `visualization`.
 
Initially I wasn't sure if it would be too much nesting (mainly wanted to try). But I think it leaves the `crabs.detector` level a bit more clear at a glance. It may help writing unit tests too.
  
 ### tracker module  
 In the `tracker` module I did some more substantial changes. I tried to follow the structure of `crabs.detector`. 
 
  - I added a `crabs.tracker.sort` module, which only has the tracker classes 
     - (somewhat equivalent to `crabs.detector.models`).
 - I added a `crabs.tracker.evaluate_tracker` module with a `TrackerEvaluate` class 
  	- (somewhat equivalent to `crabs.detector.evaluate_model` and the `DetectorEvaluate` class).
 - I added a `crabs.tracker.track_video` module which would be the main entry point 
  	- (somewhat equivalent to `crabs.detector.train_model` or `crabs.detector.evaluate_model`).
  - a `crabs.tracker.utils` submodule, which holds
     - `sort` utils: I extracted them from the sort script,
     - `tracking` utils: the previous `tracking_utils`,
     - `io` utils. This is the most substantial change. The `io` module contains what was called before `inference`, removing the class. The class was called Inference but it seemed mostly io operations (there was no prediction or tracking in the class), so I thought this format could work better.
     
### other
 I think having a separate config is a good idea!
 
 I also changed the entry points name for the tracker to something a bit more explicit (albeit longer): `detect-and-track-video`.

nikk-nikaznan:
Thank you Sofia! Looking good. I just fixed some tracking bits that throw error. I am happy with the new name convention, I just tried to follow the module name as in the issue, but have no strong opinion on either. The only thing, can we not rename `optuna_util` as `optuna`? We can either do `_optuna` or `hyperparameter_optimizer` or `hpo`?

sfmig (approved):
LGTM! 🚀 

I leave you the satisfaction of pressing the big green 🔘  😁

PR #200, opened 2024-06-27
Fix missing line

nikk-nikaznan:
I guess, when we merged 2 PRs yesterday, some lines were accidentally deleted. The one from resume training from weights only checkpoint


sfmig (approved):
🙀 thanks for spotting this!

I think we need to pass the optuna flag to the "weight-only" option too - will add and ask for your ok before squash and merging

PR #203, opened 2024-06-27
Small refactoring to load from ckpt

sfmig:
Following suggestions from a similar code review, this PR
- moves the computation of the checkpoint type to `detection_utils`,
- refactors the main training script to reduce length.

Let me know what you think!

thanks 🙌

### DIFF
--- a/crabs/detection_tracking/train_model.py
+++ b/crabs/detection_tracking/train_model.py
@@ -1,5 +1,4 @@
 import argparse
-import logging
 import os
 import sys
 from pathlib import Path
@@ -12,6 +11,7 @@ from lightning.pytorch.callbacks import ModelCheckpoint
 
 from crabs.detection_tracking.datamodules import CrabsDataModule
 from crabs.detection_tracking.detection_utils import (
+    get_checkpoint_type,
     prep_annotation_files,
     prep_img_directories,
     set_mlflow_run_name,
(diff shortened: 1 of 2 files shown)


crabs/detection_tracking/train_model.py
```diff
@@ -164,37 +165,15 @@ def core_training(self) -> lightning.Trainer:
...
-                )
-            else:
-                checkpoint_type = "weights"  # for fine tuning
-                logging.info(
-                    f"Fine-tuning training from checkpoint at: {self.checkpoint_path}"
-                )
-        else:
-            checkpoint_type = None
+        # Get checkpoint type: "full", "weights" or None
+        checkpoint_type = get_checkpoint_type(self.checkpoint_path)
```
nikk-nikaznan, line 169:
```suggestion
if self.checkpoint_path:
        checkpoint_type = get_checkpoint_type(self.checkpoint_path)
 else:
        checkpoint_type = None
```

nikk-nikaznan (approved):
Yes, make sense to me. Just small suggestion, should we only call the function if there is checkpoint_path? So we also don't need to check the condition, if the path is not there

crabs/detection_tracking/train_model.py
sfmig: yes fantastic! 🙌

nikk-nikaznan (approved):

sfmig:
By mistake I merged PR #141 before this one, instead of merging this --> rebasing PR #141 ---> merging PR #141.

Just leaving a log here in case we need to inspect the commit history at some point and are confused about it.

PR #206, opened 2024-07-02
Fix video_output error

nikk-nikaznan:
Fix video_output error if save_video is false
- Handle if save_video is false, otherwise error
- Change config to default value for save_video and save_frame_and_csv
- I thought we changed from checkpoint_path to trained_model_path, might be me accidentally reverted back, apologies

### DIFF
--- a/crabs/tracker/config/tracking_config.yaml
+++ b/crabs/tracker/config/tracking_config.yaml
@@ -4,8 +4,3 @@ score_threshold: 0.1
 max_age: 10
 # Minimum number of associated detections before track is initialised
 min_hits: 1
-# save video inference
-save_video: False
-# Save predicted tracks in VIA csv format and export corresponding frames
-# This is useful to prepare for manual labelling of tracks
-save_csv_and_frames: False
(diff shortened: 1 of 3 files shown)


sfmig (approved):
looks good!

Two small suggestions:
- I think it makes sense to just have the optional `save_video` and `save_frames`, but always output the csv.
- And should we have the output config bits as CLI instead? Since I think it will be frequently tweaked.

Let me know thoughts!

nikk-nikaznan:
By default, we always save csv. But we only save the corresponding frame when `save_frame_and_csv ` is true. It used to be in cli, and I moved it when I tried to run on cluster. A bit more neat, and I thought by default, we might just need the csv and everything else rarely? Happy to move it back to cli if you think it is easier that way.

sfmig:
Ok I see!

I like that the csv is always saved, that makes sense. Shall we call `save_frame_and_csv` --> `save_frames` then for clarity?

> we might just need the csv and everything else rarely

Yeah good point 🤔

I think as CLI arguments may be more consistent with the `detector` module. In both the training and evaluation script, we pass the mlflow folder via CLI (which is in a sense an output) and in evaluation we pass `frames_output_dir`. In training, we regulate the checkpointing behaviour via the config, but arguably that is many more parameters.

PR #207, opened 2024-07-02
adding scores

nikk-nikaznan:
Adding prediction score during tracking inference to csv file. At the moment to VIA format csv, should we have simple csv option, if we not intend to do tracking labelling?

### DIFF
--- a/crabs/tracker/utils/tracking.py
+++ b/crabs/tracker/utils/tracking.py
@@ -47,6 +47,7 @@ def write_tracked_bbox_to_csv(
     frame: np.ndarray,
     frame_name: str,
     csv_writer: Any,
+    pred_score: np.ndarray,
 ) -> None:
     """
     Write bounding box annotation to a CSV file.
@@ -62,6 +63,8 @@ def write_tracked_bbox_to_csv(
         The name of the frame.
     csv_writer : Any
         The CSV writer object to write the annotation.
+    pred_score : np.ndarray
+        The prediction score from detector.
     """
     # Bounding box geometry
     xmin, ymin, xmax, ymax, id = bbox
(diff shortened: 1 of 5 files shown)


sfmig:
I think for now we can continue with using VIA tracks format as a standard, since we also assumed the users adhere to it to analyse the bboxes tracks in [movement](https://github.com/neuroinformatics-unit/movement/pull/219)

crabs/tracker/utils/tracking.py
```diff
@@ -47,6 +47,7 @@ def write_tracked_bbox_to_csv(
     frame: np.ndarray,
     frame_name: str,
     csv_writer: Any,
+    pred_score: np.ndarray,
```
sfmig, line 50: Could `pred_score` be just a float?

(As in, do we need it to be an array?)

tests/test_unit/test_tracking_utils.py
```diff
@@ -59,7 +60,7 @@ def test_write_tracked_bbox_to_csv(csv_writer, csv_output):
         1,
         0,
         '"{""name"":""rect"",""x"":10,""y"":20,""width"":40,""height"":60}"',
-        '"{""track"":""1""}"',
+        '"{""track"":""1"", ""pred_score"":""0.9""}"',
```
sfmig, line 63:
```suggestion
        '"{""track"":""1"", ""confidence"":""0.9""}"',
```
Just because I assumed this for the movement dataloader (see example [here](https://github.com/neuroinformatics-unit/movement/blob/f8052a8151fdba6b5717b9adc695a53cdad013a1/tests/test_unit/test_load_bboxes.py#L58C26-L58C36))

sfmig (approved):
thanks loads!

A refactoring suggestion and a request to rename the output field (to amtch the movement dataloader).

crabs/tracker/utils/tracking.py
nikk-nikaznan: Yes, because we have red_score for every crabs in a frame?

PR #218, opened 2024-07-08
Fix validation and test split not being reproducible

sfmig:
### Why is this PR needed?
Currently when we create the test and validation splits we don't pass a generator. We do pass one when we create the training split.

This means that given a seed, the splitting of the dataset into `train` and `test-val` sets is reproducible, but the subsequent splitting of the `test-val` set into a `test` set and a `val` set is not.

### What does this PR do?
This PR:
- passes a new generator to the second split of the data
- adds tests for the splits generation.

Smaller bits
- adds short docstrings to existing datamodule tests.
- adds a utility function that converts a tensor of bboxes into a COCO-style dictionary for a crab dataset.

### Notes
I decided to pass a different generator for each call to `random_split` to try to make it a bit "future-proof". That way we guarantee the splits are repeatable even if some randomisation code is added in between the two calls to `random_split`.


crabs/detector/datamodules.py
```diff
@@ -170,9 +170,10 @@ def _compute_splits(
         """
 
         # Optionally fix the generator for a reproducible split of data
-        generator = None
+        generator_1, generator_2 = None, None
```
nikk-nikaznan, line 173: do we need to create two generator? I am not quite sure myself, but why we can just use the same generator for both split?

nikk-nikaznan (commented):
Well done for spotting this. I leave a comment, as I am not sure myself if we need two generators. 

Other option, can we just use [seed_everything](https://lightning.ai/docs/pytorch/stable/common/trainer.html)?

sfmig:
> Other option, can we just use [seed_everything](https://lightning.ai/docs/pytorch/stable/common/trainer.html)?

Cool, I didn't know about this!

I think for now I'd prefer to constraint the seeding to the dataset creation, because that is the part I need to be reproducible. But good to have this in the radar.

crabs/detector/datamodules.py
sfmig: I decided to use two separate generators in the end after reading online, thinking that is a bit more future proof.

crabs/detector/utils/detection.py
```diff
@@ -242,3 +243,74 @@ def slurm_logs_as_artifacts(logger: MLFlowLogger, slurm_job_id: str):
...
+    list_annotations: list[dict] = []
+    for img_id, img_bboxes in enumerate(bbox_tensors):
+        # loop thru bboxes in image
+        for bbox_row in img_bboxes:
+            x_min, y_min, x_max, y_max = bbox_row.numpy().tolist()
+            # we convert the array to list to make it JSON serializable
+
+            annotation = {
+                "id": len(list_annotations) + 1,  # 1-based
+                "image_id": img_id,
```
nikk-nikaznan, line 298: is the image_id is 0-based?

crabs/detector/utils/detection.py
```diff
@@ -242,3 +243,74 @@ def slurm_logs_as_artifacts(logger: MLFlowLogger, slurm_job_id: str):
             logger.run_id,
             f"{log_filename}.{ext}",
         )
+
+
+def bbox_tensors_to_COCO_dict(
```
nikk-nikaznan, line 248: Nice! Do you plan to use this function in the code or just for testing? If just for the test, then maybe you can move this to fixture folder for test? But I think this might

tests/test_unit/test_datamodules.py
```diff
@@ -98,6 +102,80 @@ def compare_transforms_attrs_excluding(transform1, transform2, keys_to_skip):
...
+def create_dummy_dataset():
+    """Return a factory of dummy images and annotations for testing.
+
+    The created datasets consist of N images, with a random number of bounding boxes
+    per image. The bounding boxes have fixed width and height, but their location
+    is randomized. Both images and annotations are torch tensors.
+    """
+
+    def _create_dummy_dataset(n_images):
+        # n_images = 10  # needs to be > 5 to avoid floating point errors in dataset split
```
nikk-nikaznan, line 115: Where this been assigned to 10? It is set to 5 for `test_collate_fn` and 50, 100 and 250 for `test_compute_splits`

nikk-nikaznan (approved):
Looks good to me! I think having separate generators is a good move. Nice test as well! Just some small comments below.

crabs/detector/utils/detection.py
sfmig: thanks Nik, yes I missed this! 

It is indeed 1-based by default in VIA

🦅 👁️ !

crabs/detector/utils/detection.py
sfmig: Yeah agreed! Initially I had it with the tests, but then I moved it to utils thinking that:
- It may be useful in other scenarios (for example, if we implement an option to export detections in VIA format  #231), and
- I would like to test it well at some point. I feel it is more likely to be left untested if I leave it inside tests as a helper function.

sfmig:
thanks for the help Nik! 🌟

PR #219, opened 2024-07-08
Set mlflow experiment and run name for evaluation

sfmig:
* If not passed as cli argument: experiment_name is defined from the training job's name +  '_evaluation'
* If not specified via CLI: the run name is set to be <run name of training job>_<run name of eval job>


sfmig:
Closing because combined with #220

Issue #221, opened 2024-07-11
Saving slurm logs fails if interactive job


sfmig:
If we request an interactive job, there is a job id but there are no SLURM logs, so we get an error.

Is there a way to distinguish an interactive SLURM job from a batch one?

PR #222, opened 2024-07-11
Quick fix for slurm artifacts

sfmig:
A quick and somewhat hacky fix for #221 .

We probably need to improve how the SLURM logs are added as artifacts. 
We may want to do it in the bash script instead? I opened an issue


nikk-nikaznan (approved):
make sense to me

PR #226, opened 2024-07-19
Add a README for the tracker

sfmig:
To document how we evaluate tracker performance and how we define an ID switch.


nikk-nikaznan (approved):
Looks good! Just small suggestion below. The paragraph is a bit too long, and perhaps ID switch should have its own paragraph as well?

crabs/tracker/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+To compute the total number of false negatives (or missed detections, `FN`) at a given frame `f`, we count the number of ground-truth objects that do not match with any detection at frame `f`. A ground-truth object and a detection are considered to match if their associated boxes sufficiently overlap, that is, if their intersection-over-union (IOU) is greater than a given threshold.
+
+To compute the total number of false positives (`FP`) at a given frame `f`, we count the number of detections that do not match with any of the ground-truth object defined at frame `f`.
+
+A true positive (`TP`) is defined as a detection that sufficiently overlaps with a ground-truth box (with overlap measured with the `IOU` metric). To compute the number of identity switches (`IDs`) at a given frame `f` we inspect the set of true positives, and check if for each of their ground-truth IDs, the predicted ID at frame `f` matches the predicted ID at the last frame the object was detected. If the predicted IDs do not match for the same ground-truth ID, we count that as one identity switch.
```
nikk-nikaznan, line 26:
```suggestion
A true positive (`TP`) is defined as a detection that sufficiently overlaps with a ground-truth box (with overlap measured with the `IOU` metric). 

To compute the number of identity switches (`IDs`) at a given frame `f` we inspect the set of true positives, and check if for each of their ground-truth IDs, the predicted ID at frame `f` matches the predicted ID at the last frame `f-1` the object was detected. If the predicted IDs do not match for the same ground-truth ID, we count that as one identity switch.
```

PR #242, opened 2024-11-12
Integration test for detect+track video

sfmig:
## Description

**What is this PR?**

- [ ] Bug fix
- [ ] Addition of a new feature
- [x] Other

**Why is this PR needed?**
To ensure we maintain the functionality of the detector and tracker when refactoring related parts of the codebase.

**What does this PR do?**
- Adds an integration test for the detect-and-track-video command, when ground-truth is passed.
	- Downloads sample data from GIN using pooch
	- Parametrises the test for different optional outputs (save frames or not, save tracked video or not).
- Adds a caching step to the Github actions workflow in CI
- Additionally, the CLI help for the `--output-dir` is edited, to clarify it refers to the root name of the output directory rather than the full path to it (I opened an issue #246 to support this other functionality)

### Questions
I initially was [parametrising possible flags for `--output-dir`](https://github.com/SainsburyWellcomeCentre/crabs-exploration/blob/963bec44c2262c79a4c1ca537556fe63600897d1/tests/test_integration/test_inference.py#L62) too, but decided to remove them to reduce the run time of these tests and thinking this would be better covered in a unit test. But opinions are welcome on this!

## References

For now, the integration test checks the command runs successfully and the expected outputs are created.

After reviewing and fixing the tracking evaluation, the tests should be expanded to confirm the output csv files contain the expected metrics (PR #243). 

## How has this PR been tested?

Existing and new tests pass locally and in CI.

## Does this PR require an update to the documentation?

N/a

## Checklist:

- [x] The code has been tested locally
- [x] Tests have been added to cover all new functionality
- [ n/a ] The documentation has been updated to reflect any changes
- [x] The code has been formatted with [pre-commit](https://pre-commit.com/)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,6 @@
 """Pytest configuration file."""
 
 pytest_plugins = [
+    "tests.fixtures.integration",
     "tests.fixtures.frame_extraction",
 ]
--- a/tests/fixtures/integration.py
+++ b/tests/fixtures/integration.py
@@ -9,37 +9,6 @@ import pytest
 GIN_TEST_DATA_REPO = "https://gin.g-node.org/SainsburyWellcomeCentre/crabs-exploration-test-data"
 
 
-# @pytest.fixture(autouse=True)
-# def mock_home_directory(monkeypatch: pytest.MonkeyPatch):
-#     """Monkeypatch pathlib.Path.home().
-
-#     Instead of returning the usual home path, the
-#     monkeypatched version returns the path to
-#     Path.home() / ".mock-home". This
-#     is to avoid local tests interfering with the
-#     potentially existing user data on the same machine.
-
-#     Parameters
-#     ----------
-#     monkeypatch : pytest.MonkeyPatch
-#        a monkeypatch fixture
-
-#     """
-#     # define mock home path
-#     home_path = Path.home()  # actual home path
-#     mock_home_path = home_path / ".mock-home"
-
-#     # create mock home directory if it doesn't exist
-#     if not mock_home_path.exists():
-#         mock_home_path.mkdir()
-
-#     # monkeypatch Path.home() to point to the mock home
-#     def mock_home():
-#         return mock_home_path
-
-#     monkeypatch.setattr(Path, "home", mock_home)
-
-
 @pytest.fixture(scope="session")
 def pooch_registry() -> dict:
     """Pooch registry for the test data.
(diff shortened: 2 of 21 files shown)


tests/fixtures/integration.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    # Initialise pooch registry
+    registry = pooch.create(
+        Path.home() / ".crabs-exploration-test-data",
+        base_url=f"{GIN_TEST_DATA_REPO}/raw/master/test_data",
+    )
+
+    # Download the registry file from GIN
+    # Force to download it fresh every time by using a timestamped filename
+    timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
```
sfmig, line 33: as @niksirbi made me realise, if knonw_hash = None, the file is [always downloaded](https://github.com/fatiando/pooch/blob/78cf76046b7893d757930df4dffc0f7a44bac2c0/pooch/core.py#L241).

So there is no need to set it to a different name

conftest.py
```diff
@@ -1,5 +1,6 @@
 """Pytest configuration file."""
 
 pytest_plugins = [
+    "tests.fixtures.integration",
```
niksirbi, line 4: So neat, so organised! We should definitely switch to grouping the fictures like that in `movement`.

tests/fixtures/integration.py
```diff
@@ -0,0 +1,49 @@
...
+        path=Path.home() / ".crabs-exploration-test-data",
+    )
+
+    # Load registry file onto pooch registry
+    registry.load_registry(
+        file_registry,
+    )
+
+    # Delete registry file
+    Path(file_registry).unlink()
```
niksirbi, line 47: I think this will not be needed if you switch to the-non-timestamped file name.

tests/fixtures/integration.py
```diff
@@ -0,0 +1,49 @@
...
+        url=f"{GIN_TEST_DATA_REPO}/raw/master/files-registry.txt",
+        known_hash=None,
+        fname=f"files-registry_{timestamp}.txt",
+        path=Path.home() / ".crabs-exploration-test-data",
+    )
+
+    # Load registry file onto pooch registry
+    registry.load_registry(
+        file_registry,
+    )
```
niksirbi, line 44: nitpicking: why is this command multi-line?

niksirbi (approved):
Very nice @sfmig! The integration test is very understandable and readable (even for me, having my first contact with this repo), and it makes sense.

The `pooch` bits we have already discussed intensely, and I think you can get rid of the time stamp.

FYI I timed the tests on my Macbook and they took:
- ~11 min the first time they were run (so including `pooch` download time)
- ~2.5 minutes for subsequent runs

Most time is spend on the new `test_inference.py` test.

This PR is good to go as is, but I'm wondering if you've considering running this on a smaller dataset (mostly in the interest of speeding up CI, if that's not so important ignore me).

